PR #9, opened 2015-11-10
Task/recreate subscriptions

### LINKED ISSUE #1: Recreate subscription on IoTA Restart
The subscriptions to MQTT topics are created when the devices are provisioned. They should also be recreated when the IoTA restarts.

dmoranj:
Fixes  #1


test/contextRequests/alternativeUpdate.json
```diff
@@ -0,0 +1,22 @@
...
+        {
+          "name": "temperature",
+          "type": "celsius",
+          "value": "1"
+        }
+      ]
+    }
+  ],
+  "updateAction": "UPDATE"
+}
```
gtorodelvalle: Final blank line appreciated... :)

lib/iotagent-mqtt.js
```diff
@@ -144,6 +145,24 @@ function mqttMessageHandler(topic, message) {
...
 /**
+ * Generate the list of topics related to the device, based on the device attribute definitions.
+ *
+ * @param {Object} device           Device object containing all the information about the provisioned device.
+ * @return {Array}                  List of strings names of the topics associated to the selected device.
+ */
+function generateDeviceTopics(device) {
+    var topics = [];
+
+    for (var i = 0; i < device.active.length; i++) {
```
gtorodelvalle: Does `device.active` always exist and it is an `Array` (it could be an empty one)? If `device.active` was `undefined`, this line may end up throwing an `Error`... Just in case you want to cross-check it and leave as it is right now or include an `Array,isArray()` check to be 100% sure... :) I leave it to your discretion... ;)

dmoranj: If it exists it will always be an Array (as it is returned from the same IOTA library, it would be a pretty bad bug otherwise). I'm also 98% sure it will always exist, as it comes from the Mongoose object storing the device, but I'll add that check, anyways.

gtorodelvalle:
LGTM and merging! ;) Thanks @dmoranj ;)

PR #20, opened 2015-12-22
Task/support band p1

### LINKED ISSUE #19: Support conversion of ThinkingThing attributes
Support the conversion of the P1 and B modules of Thinking Things when they come as attributes in the payload.

dmoranj:
Fixes #19


lib/thinkingThingPlugin.js
```diff
@@ -0,0 +1,165 @@
...
+            type: 'string',
+            value: fields[5]
+        });
+        result.push({
+            name: 'mode',
+            type: 'string',
+            value: fields[6]
+        });
+        result.push({
+            name: 'desconnection',
```
mrutid: disconnection instead

mrutid:
Ok

PR #28, opened 2016-01-08
FIX Agent crashes with invalid JSON

### LINKED ISSUE #27: If the given JSON is not valid, the agent crashes
If a malformed JSON is given to the agent, the parsing raises an exception and the agent crashes. This should be captured and a graceful error emitted.

dmoranj:
Fixes #27


XavierVal:
LGTM

PR #36, opened 2016-01-12
FIX Attribute alias working with timestamps

### LINKED ISSUE #35: Attribute alias don't work along with timestamp
Currently, the timestamp conversion mechanism do not work properly with aliased attributes. Tests should ensure that an attribute that is aliased to be a timestamp, will properly trigger timestamping mechanisms.

dmoranj:
Fixes #35


XavierVal:
LGTM

PR #45, opened 2016-01-19
FIX C1 should be renamed to P1

### LINKED ISSUE #44: Change C1 attribute name to P1
Change the name of the attribute generated in the device entity for a C1 device attribute to P1.

dmoranj:
Fixes #44


XavierVal:
LGTM

PR #49, opened 2016-01-19
Bug/reject configuration with resource

### LINKED ISSUE #40: MQTT Agent should reject configurations that do specify the resource
The resource value is not used in this IOTA, so a check should be performed to make sure it is allways empty (this is important, as the key to differentiate device configurations is by using the resource and apikey, so multiple different services could share the api key by generating random resources).

dmoranj:
Fixes #40


XavierVal:
LGTM

PR #54, opened 2016-01-19
Bug/mosquitto disconnection

### LINKED ISSUE #53: When there is a disconnection in the Mosquitto broker the Agent doesn't notice
When the Mosquitto shuts down or disconnects for whatever reason, once the mosquitto is up again, the subscriptions of the IOTA have been lost. It should, instead, try to recreate them when the connection is reacquired.

dmoranj:
Fixes #53


XavierVal:
LGTM

PR #62, opened 2016-01-22
Task/change configuration format

### LINKED ISSUE #59: Change the configuration format
The current configuration retrieval format returns the attributes as an array of objects with the "name" and "value" attributes. This representation is two long, and it should be changed to a format similar to the one used to report measures, that is, a plain JSON object, with each attribute being the corresponding NGSI attribute.

### LINKED ISSUE #61: Add the system time in the returned configuration
Return the system time along with all the configuration requests, in an especial attribute named "DT".

dmoranj:
Fixes #61 
Fixes #59


XavierVal:
LGTM

PR #71, opened 2016-01-27
FIX mandatory resource must be equal to the configured IOTAM resource

### LINKED ISSUE #70: Mandatory resource should be equal to the default resource in the IOTAM
When the IoT Agent registers to the IoT Agent Manager, it declares a default resource, that will be used in every group provisioning to that agent. The IOTA should enforce that default resource in the group provisioning requests, instead of enforcing the empty string.

dmoranj:
Fixes #70


AlvaroVega:
LGTM

PR #73, opened 2016-01-27
Bug/p1 and c1 with single attribute

### LINKED ISSUE #72: P1 and C1 Thinking Thing modules do not work in single attribute mode
P1 and C1 module translation was tested sending information through the multiattribute topic, but errors have been reported when the same values are sent using the single attribute one. Tests should be implemented to cover this functionality and, in case the error is reproducible, fix it.

dmoranj:
Fixes #72


anabelengp:
LGTM

PR #88, opened 2016-02-07
Add Dockerfile

drasko:
Signed-off-by: Drasko DRASKOVIC <drasko.draskovic@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+###
+# Copyright (c) Mainflux
+#
+# This file is part of iotagent-mqtt and is published under GNU Affero General Public License
+# See the included LICENSE file for more details.
+###
+
+FROM centos:6
+
+RUN yum update -y && yum install -y wget \
+  && wget http://ftp.rediris.es/mirror/fedora-epel/6/i386/epel-release-6-8.noarch.rpm && yum localinstall -y --nogpgcheck epel-release-6-8.noarch.rpm \
+  && yum install -y npm git
+
+COPY . /opt/iota-mqtt
+WORKDIR /opt/iota-mqtt
+RUN npm install
+
+ENTRYPOINT bin/iotagentMqtt.js config.js


drasko:
Please note that for successful Docker host interconnections `config.js` has to be edited appropriately (to point to the MQTT broker and Orion host).

This has been documented here: https://github.com/Mainflux/mainflux-fiware/wiki/IOTA-MQTT-Docker

The whole shebang has been tested with Moquitto MQTT breoker, also run in separate Docker container, and this `docker-compose` file: https://github.com/Mainflux/mainflux-fiware/blob/master/docker-compose.yml

dmoranj:
Sorry it's taking so long to check your PR, I'll be testing it today. I also think the Docker Compose would be a nice addition to the repository, if you wish to add it. 

Concerning the `config.js` file, in other IoTAs with Dockerfiles we have created an alternate `config-blank.js` file with those changes already done, so that the Docker image can be started with the bare repository (making it easier to upload it to Docker Hub). You can add that new file if you want, or we can just merge this Dockerfile and I'll create it in a future request.

drasko:
@dmoranj np, I can see that you are the only maintainer on several projects - it must be difficult to handle all (day is short ;)).

Anyway, if this PR is OK please integrate and I will add `config-blank.js` in subsequent PR.
I will take a look at `docker-compose` also - do you mean to use MongoDB in a composition or something else? I used IoT Agent MQTT only with in-memory storage so far.

dmoranj:
Ok. I'll merge this one and we can work on the config and compose issues in future ones.

Concerning the docker-compose, I think it's not a problem to work in memory for the docker compose, as its main use will be to have a nice and easy approach to new users (that probably would end up having their own fine-tuned configurations if they continue using the project).

drasko:
Great, thanks.

We have to think about config: can we inject it on Docker container run, or do we have to have it in the moment of iage creation? I would prefer that we can have more flexible option to pass the config in the moment of container run (so you can have a generic image on DockerHub and change the config in dependence of deployment scenario)

Also - wahat exactly components should `docker-compose` contain - additional DB, volume to back-up data from DB in container, additional MQTT broker... I think the best is that you open an issue for this task and give a few requirements, and I'll post my comments over there.

PR #115, opened 2016-05-23
FIX Different log type in libraries and general logs

dmoranj:
Fix the different log type (JSON or Pipe) for the library and the IOTA log.

### DIFF
--- a/lib/iotagent-mqtt.js
+++ b/lib/iotagent-mqtt.js
@@ -443,6 +443,7 @@ function start(newConfig, callback) {
         connectTimeout: 60 * 60 * 1000
     };
 
+    logger.format = logger.formatters.pipe;
     config = newConfig;
 
     if (config.mqtt.username && config.mqtt.password) {


lib/iotagent-mqtt.js
```diff
@@ -443,6 +443,7 @@ function start(newConfig, callback) {
         connectTimeout: 60 * 60 * 1000
     };
 
+    logger.format = logger.formatters.pipe;
```
gtorodelvalle: This will overwrite the format if set using the LOGOPS_FORMAT environment variable (used by https://www.npmjs.com/package/logops). It this what you really want? Maybe you could check if this environment variable is not set and set it to pipe only in this case (taking into consideration that from 1.. on, the default format for the logops package is json and not pipe as desired).

gtorodelvalle: Discussed offline and agreed to keep it as it is... ;)

PR #116, opened 2016-05-23
Use wildcards instead of explicit registrations for topic subscription

### LINKED ISSUE #112: Use wildcards for topics subscription instead of subscribing to each topic separately
The current code makes a specific subscription for each topic of each device. This is not needed, as a wildcard could be used to group them all.

dmoranj:
Fixes #112


AlvaroVega:
LGTM

PR #117, opened 2016-05-24
FIX Default resource location

dmoranj:
Fix the default resource location (that changed from the IoTManager object to the main iota object in last version of the library).


XavierVal:
LGTM

Issue #122, opened 2016-05-25
ADD Travis CI to the repository


dmoranj:
There is not currently a Travis CI configuration for this IOTA. It should be basically a copy from the one in [the Ultralight 2.0 IOTA](https://github.com/telefonicaid/iotagent-ul/blob/master/.travis.yml).

PR #123, opened 2016-05-25
Task/add travis ci

dmoranj:
Implements #122

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: node_js
+
+sudo: required
+
+node_js:
+ - "0.10"
+ - "4"
+
+branches:
+  only:
+    - develop
+    - master
+
+services:
+ - docker
+ - mongodb
+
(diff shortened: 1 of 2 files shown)


AlvaroVega:
LGTM

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+ - "4"
+
+branches:
+  only:
+    - develop
+    - master
+
+services:
+ - docker
+ - mongodb
```
gtorodelvalle: You're fine with running your tests against a 2.6 (I think it is) version of MongoDB, right? Since that it the one currently Travis would use according to https://github.com/travis-ci/travis-ci/issues/3694 :)

gtorodelvalle: We commented offline and decided to keep it as it is... ;)

gtorodelvalle:
LGTM

PR #145, opened 2016-06-03
Task/http binding

dmoranj:
Implements 110


lib/bindings/HTTPBinding.js
```diff
@@ -0,0 +1,283 @@
...
+    };
+
+    request(options, function(error, response, body) {
+        if (error || response.statusCode !== 200) {
+            callback(new errors.HTTPCommandResponseError(response.statusCode, error));
+        } else {
+            process.nextTick(commandHandler.updateCommand.bind(null, apiKey, device, body, callback));
+        }
+
+        callback();
```
gtorodelvalle: This `callback()` invocation may lead to unexpected behaviours as a consequence of the `callback()` being invoked twice and not with the expected arguments, doesn't it?

test/deviceProvisioning/provisionDeviceHTTP.json
```diff
@@ -0,0 +1,22 @@
...
+          "type": "celsius"
+        },
+        {
+          "name": "humidity",
+          "type": "degrees"
+        }
+      ]
+    }
+  ]
+}
```
gtorodelvalle: Final blank line appreciated... ;)

lib/bindings/HTTPBinding.js
dmoranj: Agree. Fixed in 26d89d1

gtorodelvalle:
I noticed that some of the tests files have `-test` as suffix and other ones `_test`, just in case you want to fix that... :) Everything should work regarding the automatic unit tests, though... :)

lib/bindings/HTTPBinding.js
```diff
@@ -117,9 +117,8 @@ function executeCommand(apiKey, device, serializedPayload, callback) {
             callback(new errors.HTTPCommandResponseError(response.statusCode, error));
         } else {
             process.nextTick(commandHandler.updateCommand.bind(null, apiKey, device, body, callback));
+            callback();
```
gtorodelvalle: I think you have to remove it from here too since already included into the previous function call as the callback.

dmoranj: Fixed in d8d69bd. Again

gtorodelvalle:
LGTM ;)

lib/bindings/HTTPBinding.js
dmoranj: And yet another time 64f5007. Loooong friday.

PR #206, opened 2016-11-10
Update context broker port to 1026 in stepbystep page.

pawcik:
I believe the default context broker port is 1026.


dmoranj:
It is, usually. But it's customary to use the 10026 for the unsecured Context Broker when it has a PEP in the same  machine (that will take the default port 1026). In this case, you're right, maybe the expected port would be the standard one.

dmoranj:
Thanks for your contribution.

PR #234, opened 2017-09-28
FIX shrinkwrap aspects

fgalan:
Add npm-shrinkwrap.json file in master branch and adjust .spec to take it into account at RPM generation time.

@vgarciag


vgarciag:
LGTM

Issue #222, opened 2017-03-24
Step-by-step guide not working


vears91:
Following the steps in the guide I get the following error when attempting to publish to the topic, after provisioning the device with the API

> time=2017-03-24T11:50:10.305Z | lvl=ERROR | corr=fbbbf3bc-bf3b-40a8-8d2d-e58f933c07e2 | trans=fbbbf3bc-bf3b-40a8-8d2d-e58f933c07e2 | op=IoTAgentJSON.MQTTBinding | srv=n/a | subsrv=n/a | msg=MEASURES-004: Device not found for topic [/1234/sensor01/attrs] | comp=IoTAgent

Naff16:
Im having the same problem, the values of temperature and luminosity are always empty. They not update when i execute: mosquitto_pub -t "/1234/sensor01/attrs" -m '{"l":4,"t": "31.5"}'.

vears91:
The steps in the  "Provisioning multiple devices with a Configuration" section do work for me. The steps in "Provisioning a single device with the default API Key" didn't.

Naff16:
Yes it worked the same for me, im only having  problems at 'Provisioning a single device with the default API Key'

Naff16:
In config.js file, if you remove from config.mqtt.defaultKey: '1234' and put it in config.iota.defaultKey: '1234' it will solve the problem, now i can send measures and update the device.

fgalan:
@vear91, if you confirm the solution, please close the issue. Thanks!

fgalan:
This issue seems to be related with https://github.com/telefonicaid/iotagent-json/issues/207 (maybe a duplicate?)

taliaga:
I've faced the same problem when trying to follow the step-by-step and also had to move the defaultKey configuration from the mqtt to the iota configuration sections to make the first step work.  I tend to believe it's plausible for now to fix the default config.js to stop people from attempting their first steps. 

However, I would like to better understand why this happens. A difference I've noticed in the logs of the agent with this change is that when the "defaultKey" is not in the iota section it looks there's a fallback to transport:HTTP somewhere in the chaining of agent-device. I'm afraid I'm totally new to IoTAgents and haven't yet understood the overall implementation idea so I'm still not able to offer a better suggestion :/

taliaga:
@Naff16 could you share the config file you used to successfully update the device?

fgalan:
[...]

Issue #216, opened 2017-01-17
Improvements for the documentation


dmoranj:
- Change the documentation to make it agnostic of the transport (the IoTAgent is mentioned a few times as MQTT-JSON).
- Missing HTTP transport description in installation guide (MQTT section is present). 
- Change the subtopic order in the HTTP description (Data, Commands, Configuration), as it differs from MQTT (Data, Configuraiton, Commands).
- The link https://github.com/telefonicaid/iotagent-json/blob/master/README#protocol is broken.
- https://github.com/telefonicaid/iotagent-json/blob/master/README.md#configuration has an anchor broken
- Some typos: unidirectionsl, aboveg, sepparate, gorup 
- "NGSI10/" should be changed to "/v1".
- Required Orion version should be a bit higher (as current IoTAgents are being tested with superior versions).

fgalan:
@taliaga which are the "5 items" you mention at https://github.com/telefonicaid/iotagent-json/pull/245 were solved, pls?

taliaga:
4th, 5th, 6th and 7th. For the 8th I decided to take no action as discussed in the PR. The new list would be:

- Change the documentation to make it agnostic of the transport (the IoTAgent is mentioned a few times as MQTT-JSON).
- Missing HTTP transport description in installation guide (MQTT section is present).
- Change the subtopic order in the HTTP description (Data, Commands, Configuration), as it differs from MQTT (Data, Configuraiton, Commands).
- Required Orion version should be a bit higher (as current IoTAgents are being tested with superior versions).

fgalan:
@taliaga thanks for your cleanup effort on documentation! New contributions to this issue are really welcomed :)

fgalan:
PR https://github.com/telefonicaid/iotagent-json/pull/410 improves documentation.

Not sure if this fully covers the issue, but the documentation has changed quite a bit since 17 Jan 2017 :) so I think it's better to close this old issue and, if needed, open new ones with problems in the current documentation.

Issue #207, opened 2016-11-10
[stepbystep] Device not found for topic


pawcik:
Hi,

I was following the [step by step guide](https://github.com/telefonicaid/iotagent-json/blob/master/docs/stepbystep.md) and calling:
```
mosquitto_pub -t /1234/sensor01/attrs -m '{"l":4,"t": "31.5"}'
```
I got following error in logs:
```
time=2016-11-10T14:13:31.379Z | lvl=DEBUG | corr=f6dbd06e-54f2-45d5-aa9d-73d18a505c31 | trans=f6dbd06e-54f2-45d5-aa9d-73d18a505c31 | op=IoTAgentNGSI.MongoDBGroupRegister | srv=n/a | subsrv=n/a | msg=Looking for entity params ["resource","apikey"] | comp=IoTAgent
time=2016-11-10T14:13:31.382Z | lvl=DEBUG | corr=f6dbd06e-54f2-45d5-aa9d-73d18a505c31 | trans=f6dbd06e-54f2-45d5-aa9d-73d18a505c31 | op=IoTAgentNGSI.MongoDBGroupRegister | srv=n/a | subsrv=n/a | msg=Device group for fields [["resource","apikey"]] not found: [{"apikey":"1234"}] | comp=IoTAgent
time=2016-11-10T14:13:31.382Z | lvl=ERROR | corr=f6dbd06e-54f2-45d5-aa9d-73d18a505c31 | trans=f6dbd06e-54f2-45d5-aa9d-73d18a505c31 | op=IoTAgentJSON.MQTTBinding | srv=n/a | subsrv=n/a | msg=MEASURES-004: Device not found for topic [/1234/sensor01/attrs] | comp=IoTAgent
```

After some debugging  it looks like after adding  `config.iota.defaultKey: '1234'` into `config.js` fix the problem for me. Not sure if it is the right approche.

Naff16:
Where did you put config.iota.defaultKey: '1234'???
Im adding it on config.js and the service wont start.

cumps

fgalan:
Looking to https://github.com/telefonicaid/iotagent-json/blob/master/config.js#L28, it seems it is not within config.iota, but within config.mqtt.

Not sure if some versions of IOTAgent use config.iota and other use config.mqtt for the defaultKey configuration...

fgalan:
This issue seems to be related with https://github.com/telefonicaid/iotagent-json/issues/222 (maybe a duplicate?)

[...]

PR #245, opened 2017-10-31
Fix stepbystep.md (Part 1)

### LINKED ISSUE #232: BUG: Retrieving configuration parameters from the Context Broker
To add the sleepTime attribute you must use the following NGSI request:
```
curl -X POST -H "Content-Type: application/json" -H "Accept: application/json" -H "Fiware-Service: myHome" -H "Fiware-ServicePath: /environment" -H "Cache-Control: no-cache" -d '{
"value" : "300"
}' 'http://localhost:1026/v1/contextEntities/LivingRoomSensor/attributes/sleepTime'
```
instead of 

```
curl -X POST -H "Content-Type: application/json" -H "Accept: application/json" -H "Fiware-Service: myHome" -H "Fiware-ServicePath: /environment" -H "Cache-Control: no-cache" -d '{
"value" : "300"
}' 'http://localhost:1026/v1/contextEntities/LivingRoomSensor/attrs/sleepTime'
```

[Chapter here](https://github.com/telefonicaid/iotagent-json/blob/master/docs/stepbystep.md#retrieving-configuration-parameters-from-the-context-broker)

Regards !

taliaga:
The main goal of this PR is to bring a solution to the issue #222. I.e, fixing the stepbystep, making it usable for the new users. The scope of the issue is huge because there are many things to be fixed. Thus, I've had to limit the scope of this PR to the first 2 sections of the stepbystep ("Introduction" and 
"Provisioning a single device with the default API Key").

This PR focus:
- fixes the main showstopper mentioned in #222
As bonus:
- closes #232 
- fixes 5 of the items in #216

### DIFF
--- a/config.js
+++ b/config.js
@@ -52,6 +52,7 @@ config.iota = {
     subservice: '/howto',
     providerUrl: 'http://localhost:4041',
     deviceRegistrationDuration: 'P1M',
+    defaultKey: '1234',
     defaultType: 'Thing'
 };
 
--- a/config.js
+++ b/config.js
@@ -25,6 +25,7 @@ var config = {};
 config.mqtt = {
     host: 'localhost',
     port: 1883,
+    defaultKey: '1234',
     thinkingThingsPlugin: true
 };
 
@@ -51,7 +52,6 @@ config.iota = {
     subservice: '/howto',
     providerUrl: 'http://localhost:4041',
     deviceRegistrationDuration: 'P1M',
-    defaultKey: '1234',
     defaultType: 'Thing'
 };
 
--- a/config.js
+++ b/config.js
@@ -25,7 +25,6 @@ var config = {};
 config.mqtt = {
     host: 'localhost',
     port: 1883,
-    defaultKey: '1234',
     thinkingThingsPlugin: true
 };
 
--- a/docs/stepbystep.md
+++ b/docs/stepbystep.md
@@ -8,7 +8,7 @@ Step by Step guide
 * [Provisioning multiple devices with a Configuration](#withconfiguration)
 * [Using ACLs to secure provisioning access](#acls)
 
-## <a name="introduction"/> Introduction
+## <a name="introduction">Introduction</a>
 
 This guide will show, step-by-step, how to deploy and configure an MQTT-JSON IoT Agent for its use to connect devices
 to an external NGSI Broker (aka Context Broker).
@@ -65,7 +65,7 @@ To enhance readability all the commands will be executed as root. To use other u
 and use `sudo` as usual (installation from the RPM package creates a special user for the agent, but it will not be
 used along this tutorial).
  
-## <a name="singledevice"/> Provisioning a single device with the default API Key
+## <a name="singledevice">Provisioning a single device with the default API Key</a>
 
 ### Installing the IoT Agent
 There are different ways to install the IoT Agent. In this tutorial, we will clone the last version of the agent from 
@@ -281,7 +281,7 @@ If we return now to the subscription window, we should be able to see the value 
 Along with all the information requested by the device, the IoTAgent will report the server time in the `dt` field of
 the response.
 
-## <a name="withconfiguration"/>  Provisioning multiple devices with a Configuration
+## <a name="withconfiguration">Provisioning multiple devices with a Configuration</a>
 
 In those cases where a group of devices with similar characteristics will be provisioned, a common configuration can be
 created for them. This configuration provision can be used to separate device messages between services, by 
@@ -397,7 +397,7 @@ We will get something like this:
 
 That shows the information we sent with the measures has been written to the Context Broker properly.
 
-## <a name="acls"/> Using ACLs to secure provisioning access
+## <a name="acls">Using ACLs to secure provisioning access</a>
 
 ### Overview
 The use of special APIKeys gives the IoTAgent administrator the opportunity to set different MQTT-Broker level permissions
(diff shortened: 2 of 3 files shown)


docs/stepbystep.md
```diff
@@ -8,28 +8,28 @@ Step by Step guide
 * [Provisioning multiple devices with a Configuration](#withconfiguration)
 * [Using ACLs to secure provisioning access](#acls)
 
-## <a name="introduction"/> Introduction
+## <a name="introduction">Introduction</a>
```
fgalan: Is this change fixing some issue? Or is it just stetical?

docs/stepbystep.md
```diff
@@ -52,20 +52,20 @@ information interchanged between simulated devices and the Context Broker.
 
 The following list shows the prerequisite software and versions:
 
-* Orion Context Broker (v0.26)
+* Orion Context Broker (v1.0)
```
fgalan: Is the procedure not working with v0.26?

config.js
```diff
@@ -52,6 +51,7 @@ config.iota = {
     subservice: '/howto',
     providerUrl: 'http://localhost:4041',
     deviceRegistrationDuration: 'P1M',
+    defaultKey: '1234',
```
AlvaroVega: conf.mqtt.defaultKey is referenced at https://github.com/telefonicaid/iotagent-json/blob/master/lib/iotaUtils.js#L48

docs/stepbystep.md
fgalan: I mean, I'm not a fan of so ancient veresion :), but we should have a good reason to change

taliaga: Oh sure, neither do I, but yes. It was causing the whole text of the section to be part of the link (checkout http://fiware-iotagent-json.readthedocs.io/en/latest/stepbystep/index.html)

docs/stepbystep.md
taliaga: This is to raise the discussion mentioned in #216. I'd keep it as it is in lack of responses, but since I was taking the chance to update this file...

I honestly haven't tested other versions than the latest, so who maybe know this is @dmoranj. I coudn't spot any hint of the correct version by looking in this repo, maybe a different one has been used for integration tests that I'm not aware of?

taliaga:
Also very important, the "**_Retrieving configuration parameters from the Context Broker_**" part of the step-by-step, which would be the last thing in the "**_Provisioning a single device with the default API Key_**" part, is neither working form me. Is that working for any of you guys? Maybe there's something that still needs to be fixed.
I've tried a thing or two (like using the "configuration" type for the sleepTime) but with no success so far. Any idea?

config.js
taliaga: Right, maybe this explains the mystery of the missing test in CI. In any case I still don't understand why failure is on AMPQ only and not the others.

Maybe I need to keep it there in "mqtt" also, but note that if kept in both configs (mqtt and iota) we're back to the first problem 😕 . This means the workaround mentioned in #207 is not the way to go.

I may need some perspective here. If you have any suggestion please let me know.

AlvaroVega: After fix this issue: https://github.com/telefonicaid/iotagent-json/issues/247 you will see a error trace about defaultKey is not fund. Could you please try it again?

docs/stepbystep.md
fgalan: I see

NTC

docs/stepbystep.md
fgalan: Maybe a solution is to consider both, I mean:

```
* Orion Context Broker (we have tested this procedure with v0.26 and v1.0)
```

or a similar wording.

taliaga: I'll leave it as it was, the change would not add any value.

config.js
taliaga: Sorry, I can't see how that fix and what you are asking connects to the problem being discussed. I'd be glad if you could help me understand.

Let's recap for a bit. The problem is, using a `config.js` as it is in the root of the project will not produce the expected results of the step-by-step (section "Provisioning a single device with the default API Key"). In particular, attribute values are not seen.

Maybe there's nothing wrong with the agent config, and what needs to be fixed is the the doc. I'm assuming the guide is right and the config is wrong, but please let me know if you think it's otherwise!

The config.js has defaultKey: '1234' only in the config.mqtt section.
People reported that the "workaround" was to move that definition to the config.iota section.
The workaround seems to be wrong for some tests in CI.

My question is, do you know which is the correct location of such a config?. 

Trying to answer your question, I tried the latest version (docker image tagged latest from 5 hours ago which I suppose brings the latest fix).
- Keeping the config.js as it is, it still "does not work".
- I removed all the definitions of defaultKey from the config.js being used to see if I got that message you mention with the fix but I didn't see anything (using loglevel DEBUG) :s.

fgalan: In order not blocking the PR, an option would be to leave the changes in config.js out and, once they get clarified (using https://github.com/telefonicaid/iotagent-json/issues/222 to hold dicussion), do it in a next PR.

What do you think?

taliaga: Sounds good to me :)

fgalan: Thus, NTC

docs/stepbystep.md
fgalan: NTC

fgalan:
LGTM

Passing the ball to @AlvaroVega for final LGTM + merge

fgalan:
@taliaga thanks for the contribution!

AlvaroVega:
LGTM

PR #288, opened 2018-06-15
update dockerfile to mount a local configuration file

mblaettler:
Move the config.js file in a separate config directory. This makes
it possible to mount a local configuration file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,10 @@ RUN yum update -y && \
   # This cannot be removed using yum as yum uses hard dependencies and doing so will uninstall essential packages
   rm -rf /usr/share/i18n /usr/{lib,lib64}/gconv && \
   # Don't need old log files inside docker images
-  rm -f /var/log/*log
+  rm -f /var/log/*log && \
+  # Create config directory and populate it with the default configuration
+  mkdir /opt/iotajson/config && \
+  mv /opt/iotajson/config.js /opt/iotajson/config/config.js
 
-ENTRYPOINT bin/iotagent-json config.js
+ENTRYPOINT bin/iotagent-json config/config.js
 
--- a/docs/installationguide.md
+++ b/docs/installationguide.md
@@ -38,6 +38,19 @@ The IoTA will then be installed as a linux service, and can ve started with the 
 ```
 service iotaJSON start
 ```
+
+#### Using Docker
+A docker container is available on docker hub. It will start the container with the default settings defined
+in `config.js`.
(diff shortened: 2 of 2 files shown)


AlvaroVega:
I think this kind of feature should be optional (activated by a var en Dockerfile) an disabled by default.

mblaettler:
It is optional. If no volume is mounted (no `-v` flag passed), nothing will change. The default configuration, which is part of the repository, will be used. It's the same behaviour in other containers (e.g. [mosquitto](https://github.com/eclipse/mosquitto/tree/master/docker/1.4.12) or [influxdb](https://github.com/influxdata/influxdata-docker/tree/master/influxdb/1.4))

I can clarify this in the documentation if you'd like.

mblaettler:
I updated the documentation to clarify that the configuration is optional.

fgalan:
Not sure if I'm understanding this fully and correctly but I think the solution should be focused in enhance with new variables `cygnus-entrypoint.sh`, intead that touching Dockefile.

For instance, let's consider:

```
cygnus-ngsi.sources.http-source.port = 5050 
```

A env var should be defined for that (e.g. CYGNUS_HTTP_SOURCE_PORT although @AlvaroVega could provide an alternative name if that is not correct).

Then, `cygnus-entrypoint.sh` would add a check for that variable and adjust the resulting agents.conf file properly, e.g:

```
    if [ "$CYGNUS_HTTP_SOURCE_PORT" != "" ]; then
        sed -i '/#'${CYGNUS_AGENT_NAME}'.sources.http-source.port/c '${CYGNUS_AGENT_NAME}'.sources.http-source.port = '${CYGNUS_HTTP_SOURCE_PORT} ${FLUME_HOME}/conf/${AGENT_CONF_FILE}
fi
```

@AlvaroVega could provide aditional info if needed (e.g. documentation aspects, if any) or if I'm totatlly wrong :)

fgalan:
(Ignore my last comment -now deleted. It was related with another issue and it has been moved)

dcalvoalonso:
I think that this PR is already covered by the recently merged https://github.com/telefonicaid/iotagent-json/pull/337

What do you think?

fgalan:
I agree.

@mblaettler what do you think? Can we close this PR?

mblaettler:
Looks good. Although I'm not sure if the solution provided in #337 could also be applied to a kubernetes deployment specification (with using a config-map).

But in my opinion, the issue can be closed.

PR #304, opened 2018-07-27
upgrade mqtt dep to 2.18.3

AlvaroVega:


package.json
```diff
@@ -54,7 +54,7 @@
     "dateformat": "1.0.12",
     "iotagent-node-lib": "git://github.com/telefonicaid/iotagent-node-lib.git#master",
     "logops": "1.0.0-alpha.7",
-    "mqtt": "1.14.1",
+    "mqtt": "2.18.3",
```
fgalan: The entry in CNR:

```
Fix: upgrade mqtt dep from 1.7.0 to 1.14.1
```

Should be updated.

AlvaroVega: I think we can postpone this upgrade for a further version

fgalan: Now that a new release has been just released, this PR could be merged. First the following entry should be added to CNR:

```
Fix: upgrade mqtt dep from 1.14.1 to 2.18.3
```

PR #324, opened 2018-09-11
ensure qos is int

AlvaroVega:


lib/configService.js
```diff
@@ -108,7 +108,7 @@ function processEnvironmentVariables() {
     }
 
     if (process.env.IOTA_MQTT_QOS) {
-        config.mqtt.qos = process.env.IOTA_MQTT_QOS;
+        config.mqtt.qos = parseInt(process.env.IOTA_MQTT_QOS);
```
fgalan: What would happen if process.env.IOTA_MQTT_QOS is not an integer? Would parseInt() raises an exeception and the IoTAgent exits in a proper manner?

AlvaroVega: String '1' is translated to integer=1

If IOTA_MQTT_QOS is not and integer then parseInt do not raises an exception, just return NaN

AlvaroVega: In that case NaN in QoS is interpreted as QoS 0

fgalan: What about if the configuration comes from config.js file? It seems this parseInt() works only when the configuraiton is provided by env var.

Maybe it is better to do the check after env var has been processed, in the resulting config.mqtt.qos variable.

AlvaroVega: Where can we check that? If I'm not wrong other config values are not cheched in that sense.

AlvaroVega: Here https://github.com/telefonicaid/iotagent-json/blob/master/lib/bindings/MQTTBinding.js#L93 should be tested

fgalan: Yes... probably that point would ensure that QoS is correctly processed, not matter if the configuration was provided by config.js or by env vars. I'd suggest to move the check there.

PR #339, opened 2018-10-27
Updates dependencies and supports Node.js v8

dcalvoalonso:
Following what was discussed in https://github.com/telefonicaid/iotagent-node-lib/issues/681 and the example of https://github.com/telefonicaid/iotagent-node-lib/pull/682, this PR aims to update all dependencies, supporting newer versions of Node.js and removing the use of Grunt.


fgalan:
Is this PR ready to be reviewed? I have seen that "WIP" was removed in the name and that travis got ok in the last commit, but I'm asking to be fully sure no new commit is expected :)

dcalvoalonso:
Yes @fgalan! I think it is ready to get your comments!

test/subscriptions/subscriptionRequest.json
```diff
@@ -16,5 +16,6 @@
         "warningLevel"
       ]
     }
-  ]
+  ],
+  "attributes":["sleepTime","warningLevel"]
```
fgalan: Why this change was needed? Were test not passing without the change?

test/unit/ngsiv2/contextRequests/TTModuleBLong.json
```diff
@@ -1,4 +1,8 @@
 {
+  "humidity":{
```
fgalan: Why this change was needed? Were test not passing without the change?

test/unit/ngsiv2/configuration-api_test.js
```diff
@@ -70,7 +71,7 @@ describe('Configuration API support', function() {
 
     beforeEach(function(done) {
         nock.cleanAll();
-
+        originalResource = config.iota.defaultResource;
```
fgalan: Could you elaborate on this change, pls?

package.json
```diff
@@ -19,43 +19,34 @@
...
   "scripts": {
-    "test": "grunt test"
+    "clean": "rm -rf package-lock.json && rm -rf node_modules && rm -rf coverage",
+    "test": "mocha --recursive 'test/**/*.js' --reporter spec --timeout 3000 --ui bdd --exit",
+    "test:watch": "npm run test -- -w ./lib",
+    "lint": "jshint lib/**/*.js test/**/*.js --config .jshintrc",
+    "test:coverage": "istanbul cover _mocha -- --recursive 'test/**/*.js' --reporter spec --exit",
+    "watch": "watch 'npm test && npm run lint' ./lib ./test"
   },
   "devDependencies": {
```
fgalan: As done in PR https://github.com/telefonicaid/iotagent-node-lib/pull/682, this PR should include a similar description of dependencies upgrade and removed dependencies in CHANGES_NEXT_RELEASE file

docs/usermanual.md
```diff
@@ -370,128 +370,74 @@ of this attribute, adding the "batteryType" and "percentage" fields to the entit
...
 # Use git-bash on Windows
-grunt coverage-report
+npm run test:coverage
 ```
 
+### Clean
 
-### Code complexity
-Plato
+Removes `node_modules` and `coverage` folders, and  `package-lock-json` file so that a fresh copy of the project is restored.
```
fgalan: package-lock-json -> package-lock.json

lib/bindings/AMQPBinding.js
```diff
@@ -135,17 +135,24 @@ function start(callback) {
                     }
                 });
                 conn.on('close', function() {
-                    config.getLogger().error(context, 'reconnecting');
-                    if (numRetried <= retries) {
-                        numRetried++;
-                        return setTimeout(createConnection,
-                                          retryTime * 1000,
-                                          callback);
+                    // If amqpConn is null, the connection has been closed on purpose
```
fgalan: As far as I understand, this change is not due to changes in the amqp library API, as it has not been updated in this PR (https://github.com/telefonicaid/iotagent-json/pull/339/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L51).

Thus, could you elaborate on this change, pls?

lib/bindings/AMQPBinding.js
```diff
@@ -173,8 +180,10 @@ function start(callback) {
 
     function assertQueue(callback) {
         config.getLogger().debug(context, 'asserting queues');
-        amqpChannel.assertQueue(queue, {exclusive: false}, callback);
-        amqpChannel.assertQueue(queue + '_commands', {exclusive: false}, callback);
+        amqpChannel.assertQueue(queue, {exclusive: false}, function () {
```
fgalan: ...and this other one, pls.

package.json
```diff
@@ -19,43 +19,34 @@
     "node": ">=4.8.4"
   },
   "scripts": {
-    "test": "grunt test"
+    "clean": "rm -rf package-lock.json && rm -rf node_modules && rm -rf coverage",
+    "test": "mocha --recursive 'test/**/*.js' --reporter spec --timeout 3000 --ui bdd --exit",
+    "test:watch": "npm run test -- -w ./lib",
+    "lint": "jshint lib/**/*.js test/**/*.js --config .jshintrc",
```
fgalan: I'd suggest using:

```
     "lint": "jshint lib/ --config .jshintrc && jshint test/ --config test/.jshintrc", 
```

Due to:

1. We have two `.jshintrc` to have the possibility of using a different style for regular code and test (although maybe currently we are not using that and both `.jshintrc` are basically the same).
2. It seems the `lib/**/*.js` glob doesn't cover the content on `lib/` directory itself. It seems that `lib/` works in a recursive way and doesn't have that problem. It is easy to check: remove a `;` in `lib/constant.js` and you will see that current approach doesn't detect while my above proposal does.

test/unit/commandsAmqp-test.js
```diff
@@ -182,7 +182,7 @@ describe('AMQP Transport binding: commands', function() {
             setTimeout(function() {
                 contextBrokerMock.done();
                 done();
-            }, 200);
+            }, 1000);
```
fgalan: Has this timeout being increated to reduce the probability of fail in the "glitching AMQP" test at travis CI?

fgalan:
@dcalvoalonso a review has been done an a few round of comment provided. Thanks!

test/subscriptions/subscriptionRequest.json
dcalvoalonso: The update of Nock module implies that payloads must have exact matching. This was also reported in a previous PR of iotagent-node-lib: https://github.com/telefonicaid/iotagent-node-lib/issues/660#issuecomment-421041564
So in order to pass this test, we needed to complete the nock payload.

test/unit/ngsiv2/contextRequests/TTModuleBLong.json
dcalvoalonso: Same reason explained in https://github.com/telefonicaid/iotagent-json/pull/339#discussion_r230902506

test/unit/commandsAmqp-test.js
dcalvoalonso: Right, I think that maybe the glitching problem was caused because of a very short timeout. This already happened in some unit tests of the LWM2M agent. Let's see if this increase helps to reduce the frequency of this issue and if not we can revert it easily.

package.json
dcalvoalonso: Fixed with ca5f3959de06ad0a2bd8bdd6485d236169fa98d3

docs/usermanual.md
dcalvoalonso: Fixed with 31a12f817c38b78879a29822ad5e5a32f83d6a9d

test/unit/ngsiv2/configuration-api_test.js
dcalvoalonso: This is needed due to the removal of `grunt-mocha-test` to execute the tests. It seems that mocha caches modules and files loaded with `require`.
From https://mochajs.org/api/mocha:
> To run tests multiple times (or to run tests in files that are already in the require cache), make sure to clear them from the cache
There are many other references the Internet and the solution seems to be to delete the cache: https://stackoverflow.com/questions/22975864/how-to-require-same-file-in-mocha-test

In our case, this is a problem for the configuration file since we use always the same in all the tests and it is loaded using:
```
config = require('./config-test.js'),
```

Therefore, the file is actually only loaded in the first test. In the following ones, it is used the cached version. The problem happens when the configuration is modified by one test, as happens in https://github.com/telefonicaid/iotagent-json/blob/master/test/unit/ngsiv2/configuration-api_test.js#87. If this change is not reverted, the next tests will not have the appropriate value in ` config.iota.defaultResource`.

`grunt-mocha-test` may take care of this kind of things, probably in https://github.com/pghalliday/grunt-mocha-test/blob/master/tasks/lib/MochaWrapper.js#36.

Nevertheless, you can check that tests fail for the master branch if you use directly mocha:

```
node_modules/mocha/bin/_mocha tools/mocha-globals.js test/unit/ngsiv2/configuration-api_test.js test/unit/ngsiv2/HTTP_reveice_measures-test.js --exit
```

lib/bindings/AMQPBinding.js
dcalvoalonso: After updating `async` module, tests related to AMQP binding started to fail due to a `Callback already called error`.

I checked the code of https://github.com/telefonicaid/iotagent-json/blob/master/lib/bindings/AMQPBinding.js and identify some reasons:
- In https://github.com/telefonicaid/iotagent-json/blob/master/lib/bindings/AMQPBinding.js#L176 and https://github.com/telefonicaid/iotagent-json/blob/master/lib/bindings/AMQPBinding.js#L177, the same `callback` is passed to the two `assertQueue` functions. This will cause that both of them call to the same `callback` function...
- In https://github.com/telefonicaid/iotagent-json/blob/master/lib/bindings/AMQPBinding.js#137, if the connection closes, this part of the code tries to reconnect using return `setTimeout(createConnection,retryTime * 1000,callback)`. As you can see, we are passing again the callback to setTimeout, but this callback was already called when the connection was initially established (https://github.com/telefonicaid/iotagent-json/blob/master/lib/bindings/AMQPBinding.js#L148). Then, if the reconnection succeeds, we will have a `Callback already called` error. For this reason, I use directly return `setTimeout(createConnection, retryTime * 1000);` without callback.
- Finally, https://github.com/telefonicaid/iotagent-json/blob/master/lib/bindings/AMQPBinding.js will be executed also when the connection is closed by `close` function of https://github.com/telefonicaid/iotagent-json/blob/master/lib/bindings/AMQPBinding.js#L209, and the code will try to reconnect... I think that this does not make sense. For that reason, in `close` function I set amqpConn and check it in `close` event handler.

lib/bindings/AMQPBinding.js
dcalvoalonso: All AMQP related changes are explained in https://github.com/telefonicaid/iotagent-json/pull/339#discussion_r231403078

lib/bindings/AMQPBinding.js
fgalan: Thank you for so detailed explanation :)

Everything seems to be ok. My only doubt if this is could be considered an existing issue in the AMQP binding logic (that the update of aync has raised but that has been always there) so the fix should be mentioned in the CNR. Something like:

```
Fix: AMQP callback over-calling
```

or maybe something more precise.

Up to you :) Either close this comment thread with NTC (if CNR is ok as it is now) or with Fixed in (if CNR needs to be modified)

lib/bindings/AMQPBinding.js
fgalan: NTC (let's move dicussing to the cited comment)

test/unit/ngsiv2/configuration-api_test.js
fgalan: Thanks for the explation. If I'm understanding correctly, removing grunt-mocha makes a little more complated developing the test, as each test needs to have care of levaing the config object as it is found at origin.

However, I think is a reasonable price to pay. Thus, NTC

lib/bindings/AMQPBinding.js
dcalvoalonso: Fixed with 9d73a9bc73942890c0e31901649b01d61b581deb

AlvaroVega:
LGTM

PR #340, opened 2018-10-29
use group timestamp conf

AlvaroVega:
Merge after PR https://github.com/telefonicaid/iotagent-node-lib/pull/669

### DIFF
--- a/lib/iotaUtils.js
+++ b/lib/iotaUtils.js
@@ -135,6 +135,9 @@ function findOrCreate(deviceId, transport, group, callback) {
             if (!newDevice.transport) {
                 newDevice.transport = transport;
             }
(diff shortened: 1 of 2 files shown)


lib/iotaUtils.js
```diff
@@ -135,6 +135,9 @@ function findOrCreate(deviceId, transport, group, callback) {
             if (!newDevice.transport) {
```
fgalan: CNR entry would be needed?

AlvaroVega: Fixed in https://github.com/telefonicaid/iotagent-json/pull/340/commits/e1556671e5b766dcd6a4309ecc11277088b2554e

PR #344, opened 2018-11-01
Minor Dockerfile fix

jason-fox:
* Save `Dockerfile` as UTF-8 - to correct the `MAINTAINER`  Telefónica I+D
* Add `.dockerignore` file
* Use an `ARG` to set the node version

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,5 @@ coverage/
 docs/
 ghpages/
 test/
-tools/
 rpm/
 README.md
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,5 @@ docs/
 ghpages/
 test/
 tools/
+rpm/
+README.md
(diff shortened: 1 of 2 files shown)


.dockerignore
```diff
@@ -0,0 +1,5 @@
+coverage/
+docs/
+ghpages/
+test/
+tools/
```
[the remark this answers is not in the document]
jason-fox: Fixed - 97d940d

.dockerignore
```diff
@@ -0,0 +1,7 @@
...
```
dcalvoalonso: ´This folder will be removed with https://github.com/telefonicaid/iotagent-json/pull/339 so I would suggest not to include it.

jason-fox: Fixed - 5bb8375 - `tools/` removed.

dcalvoalonso:
LGTM

PR #347, opened 2018-11-07
Remove `##Index` from ToC since it is not required

jason-fox:
* The ToC is suppressed with the ReadtheDocs CSS
* The ToC Header cannot be suppressed without CSS3

Therefore the title of the ToC should be removed to avoid confusion

* Add `bash` code hints where not found

Readthedocs can mis-diagnose code styles sometimes.

Related to telefonicaid/iotagent-node-lib/issues/648


dcalvoalonso:
LGTM

PR #363, opened 2019-01-16
Timeout for HTTP commands

### LINKED ISSUE #152: No timeout defined for HTTP commands
Copied from https://github.com/telefonicaid/iotagent-node-lib/issues/373.

It should be defined a timeout for commands responses.

Also it would be nice to change the status of the command in case the commnad endpoint does not respond

Jagatjot:
Fixes #152

### DIFF
--- a/config.js
+++ b/config.js
@@ -87,6 +87,20 @@ config.amqp = {
     options: {durable: true}
 };
 
+/**
+ * Configuration for the HTTP transport binding.
+ */
+config.http = {
+    /**
+     * South Port where the Ultralight transport binding for HTTP will be listening for device requests.
+     */
+    port: 7896,
+    /**
+     * HTTP Timeout for the http command endpoint (in miliseconds).
+     */
+    //timeout: 1000
+};
+
 config.iota = {
     /**
      * Configures the log level. Appropriate values are: FATAL, ERROR, INFO, WARN and DEBUG.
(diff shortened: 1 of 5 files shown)


config.js
```diff
@@ -87,6 +87,20 @@ config.amqp = {
...
 };
 
+/**
+ * Configuration for the HTTP transport binding.
+ */
+config.http = {
+    /**
+     * South Port where the Ultralight transport binding for HTTP will be listening for device requests.
+     */
+    port: 7896,
```
fgalan: From where this default comes? Is the same that used by default if no configuration is provide (i.e. as it is currently at master branch)?

Jagatjot: I could not exactly understood what you wanted to ask ? Default in terms of what ?

fgalan: I'l try to make more explicit my question...

If IOTA-UL would run without specifying config.http.port (i.e. as config.js is now on master branch, where that element is not included), which port will be used?

Jagatjot: You mean to say if we remove http port from config.js then which port it will use by default. Is my understanding correct ?

fgalan: Yes, it is.

Bottom line is that it is unclear where the 7896 number comes from. Does it come from somewhere? Or did you pick it at random?

Jagatjot: The default HTTP port for IOTA is 7896. It is already mentioned in the docker-compose file of iotagent-json as an environment variable. You can verify it in the link below- https://github.com/telefonicaid/iotagent-json/blob/53c7d99cb680a123ec30152d0dae7674be724880/docker/README.md

Jagatjot: The current config.js file doesn't contain any configuration for http as it is in master which on starting the iotagent gives a configuration error as shown in the screenshot. 
without_http . But when we define http configuration as done in 
e2b9937 , the configuration error is also resolved. Thank you!

fgalan: Yes, it seem to be everywhere :) https://github.com/telefonicaid/iotagent-json/search?q=7896&unscoped_q=7896

NTC

dcalvoalonso:
LGTM

PR #370, opened 2019-01-18
Feature/add coveralls

jason-fox:


package.json
```diff
@@ -39,11 +39,12 @@
   },
   "keywords": [],
   "dependencies": {
-    "express": "~4.16.4",
     "amqplib": "~0.5.1",
     "async": "2.6.1",
     "body-parser": "1.18.3",
+    "coveralls": "^3.0.2",
```
fgalan: We try to avoid `^` usage in packages.json dependencies (we have had bad experience with it in the past). Could you change it to `~`, please?

In addition, I think this would be better placed inside `devDependencies`.

jason-fox: Fixed 29ccb7ce

PR #372, opened 2019-01-18
Improve MQTT client configurability (adding at least support for mqtts)

aarranz:
This PR pretends to add support for using the mqtts transport. This PR also provides support for using client authentication using certificates and allows to configure whether or not allowing invalid (e.g. expired, self-signed, ...) certificates through the `rejectUnauthorized` option.

I have being using these changes with success, but it is true that this PR is currently not providing any unit tests, documentation... and probably it is not using the proper code style. Please, let me know whatever change is required to make this acceptable for inclusion 😃.

### DIFF
--- a/lib/bindings/MQTTBinding.js
+++ b/lib/bindings/MQTTBinding.js
@@ -123,21 +123,23 @@ function unsubscribeAll(callback) {
  * Start the binding.
  */
 function start(callback) {
+    let configuration = config.getConfig().mqtt;
     var options = {
+        protocol: configuration.protocol != null ? configuration.protocol : "mqtt",
+        host: configuration.host != null ? configuration.host : "localhost",
+        port: configuration.port != null ? configuration.port : 1883,
+        key: configuration.key != null ? configuration.key : null,
+        ca: configuration.ca != null ? configuration.ca : null,
+        cert: configuration.cert != null ? configuration.cert : null,
+        rejectUnauthorized: configuration.rejectUnauthorized != null ? configuration.rejectUnauthorized : true,
+        username: configuration.username != null ? configuration.username : null,
+        password: configuration.password != null ? configuration.password : null,
         keepalive: 0,
         connectTimeout: 60 * 60 * 1000
     };
 
-    if (config.getConfig().mqtt.username && config.getConfig().mqtt.password) {
-        options.username = config.getConfig().mqtt.username;
-        options.password = config.getConfig().mqtt.password;
-    }
-
-    mqttClient = mqtt.connect(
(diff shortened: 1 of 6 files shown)


lib/bindings/MQTTBinding.js
```diff
@@ -123,21 +123,23 @@ function unsubscribeAll(callback) {
  * Start the binding.
```
fgalan: An entry at CHANGES_NEXT_RELEASE describing the change introduced by this PR should be added.

lib/bindings/MQTTBinding.js
```diff
@@ -123,21 +123,23 @@ function unsubscribeAll(callback) {
...
  */
 function start(callback) {
+    let configuration = config.getConfig().mqtt;
     var options = {
+        protocol: configuration.protocol != null ? configuration.protocol : "mqtt",
```
fgalan: New parameters should:

* Be included at https://github.com/telefonicaid/iotagent-json/tree/master/docs/installationguide.md#mqtt-configuration
* Have an associated environment variable (IOTA_MQTT_xxxx), documented also in that file and properly treated in the code.

fgalan:
In general I think the PR is ok. Thanks for the contribution!

However, some things need to be fixes before merging. Please find my comment in "Diff tab". Apart form them:

* Travis is currently failing in the style check step. Please have a look to the travis log
* With regards to tests, please have a look to MQTT*.js files under https://github.com/telefonicaid/iotagent-json/tree/master/test/unit. I think it doesn't worth the effort to have a paralel test in MQTTS for each one in MQTT but including 1-2 basic MQTTS tests would be a good idea.

fgalan:
In addition, it would be great to have also a similar PR for IOTA-UL agent to keep their features in sync.

aarranz:
Ok @fgalan, I'm working on it 😄. Regarding IOTA-UL, code should be more or less the same, so I will finish first this and once approved I will translate it into IOTA-UL.

aarranz:
I've updated the code and it is ready for a new review.

When running Travis, the "should publish the command information in the AMQP topic" test is failing (for some of the jobs, it seems a random failure), it is not failing in my working copy... (any tip?)

package.json
```diff
@@ -32,6 +32,8 @@
     "nock": "10.0.1",
     "mocha": "5.2.0",
     "should": "13.2.3",
+    "sinon": "~6.1.0",
```
fgalan: Mention the addition of these modules in CHANGES_NEXT_RELEASE, pls.

fgalan: Discussed at skype, maybe sinon-chai dependeny could be avoided.

aarranz:
Fixed in 71e7bd304a0d4bee89a65b5e903d099bc6ea44aa

PR #376, opened 2019-01-28
FIX broken unit test due to mock failing with ?type expecations

fgalan:
Please @dcalvoalonso have a look.


dcalvoalonso:
Maybe we should add a CNR entry?

dcalvoalonso:
For the rest, LGTM

fgalan:
> Maybe we should add a CNR entry?

This is a new feature in the library, so the CNR entry should be (and actually is) there, see
https://github.com/telefonicaid/iotagent-node-lib/blob/master/CHANGES_NEXT_RELEASE#L5

dcalvoalonso:
Ok, NTC then.

Issue #342, opened 2018-10-31
Add a second Dockerfile for Automated Builds


jason-fox:
@dcalvoalonso @fgalan - Something for discussion, can we offer two docker files in the repo to improve the user experience?

## Current situation

The Docker image for each of the IoT Agents is a **public automated build**. This means that:

1) The description of the Image comes from the `README` in the same directory as the `Dockerfile` 
2) The `latest` image will be built every time a commit occurs - i.e. `latest` is bleeding edge development

## What is the issue

There are two problems with this.

1) The full description is currently a copy of the GitHub readme, which is not very useful for users and contains broken links, random HTML etc: https://hub.docker.com/r/fiware/iotagent-json/

2) The `latest` tag is very unstable and potentially broken, untested etc (yes I know using `latest` is bad practice)

## What I think needs to be done:

Since you can't use `COPY` from the parent directory, the current `Dockerfile` must lie in root. However there are three standard methods for getting source in a release. You can use `COPY`, `git clone` or `wget` a zip file. `git clone` is overkill for a slim release build, but `wget` of a labelled release can be easily used  to get stable source code.

I have recently fixed this in the Tutorials : https://github.com/Fiware/tutorials.Step-by-Step as follows:

The **root** contains the equivalent of the existing set-up for developers to play with:
*  a file called `Dockerfile-sample` .  - this loads the source code using a `COPY . <destination>` command
*  a `.dockerignore` file - this filters out the unnecessary files

This is effectively just a developer's hacker file - you can make changes, rebuild or try different base images.

There is a separate subdirectory called docker which holds two more files:

*  a file called `Dockerfile` .  - this is almost a duplicate of the above, but  it obtains the latest Release from GitHub using `wget` 
*  a `README` file - this forms the Full description for the Automated build. - A useful **Docker Hub** `README` would be something like - https://fiware-iotagent-json.readthedocs.io/en/latest/installationguide/index.html#configuration-with-environment-variables

This is the only file called `Dockerfile` in the whole repository and hence the one used for the automated builds.

## What is the difference?

The difference between the two `Dockerfiles` is as shown:

###  Dummy `Dockerfile-sample` in ROOT

```docker 
COPY <source> <destination> 
```

### Actual `Dockerfile`  with the `README`

```docker
RUN RELEASE=$(curl -s https://api.github.com/repos/${OWNER}/${REPOSITORY}/releases/latest | grep 'tag_name' | cut -d\" -f4) && \
[...]

PR #380, opened 2019-02-01
Entry.sh file for username, password authentication in docker

### LINKED ISSUE #366: [MUST] Dockerization must protect Username/Password ENV
With the addition of Oauth2 support authentication with  using FIWARE Keyrock, usernames and passwords are now part of the Docker environment variables. Currently these can only be passed using plain text. The addition of **Docker Secrets** support protecting these passwords is necessary to plug a potential security flaw.

This is a simple addition of a script   in the same manner as PostGres  (and relevant documentation of course) Cygnus does it already. This flaw relates to all IoT Agents.

https://github.com/telefonicaid/iotagent-node-lib/pull/726#discussion_r242214046

Within the Docker container, Passwords etc can only be passed protected by Docker Secrets
- MUST  requirement from TSC

Jagatjot:
Fixes #366

### DIFF
--- a/config.js
+++ b/config.js
@@ -135,11 +135,14 @@ config.iota = {
         port: 4041
     },
     authentication: {
-        enabled: true,
+        enabled: false,
         type: 'keystone',
         header: 'X-Auth-Token',
         host: 'localhost',
         port: '5000',
+     /**
+      * Usename and password must be overriden by setting them using difficult pattern for security issues.
+      */
         user: 'iotagent',
         password: 'iotagent'
     },
--- a/config.js
+++ b/config.js
@@ -140,11 +140,8 @@ config.iota = {
         header: 'X-Auth-Token',
         host: 'localhost',
         port: '5000',
-     /**
-      * Usename and password must be overriden by setting them using difficult pattern for security issues.
-      */
-        user: 'iotagent',
-        password: 'iotagent'
+        user: 'IOTA_AUTH_USER',
+        password: 'IOTA_AUTH_PASSWORD'
     },
     /**
      * Defines the configuration for the Device Registry, where all the information about devices and configuration
--- a/config.js
+++ b/config.js
@@ -134,15 +134,44 @@ config.iota = {
      */
         port: 4041
     },
-    authentication: {
-        enabled: false,
-        type: 'keystone',
-        header: 'X-Auth-Token',
-        host: 'localhost',
-        port: '5000',
-        user: 'IOTA_AUTH_USER',
-        password: 'IOTA_AUTH_PASSWORD'
-    },
+   
+    /**
+     * Configuration for secured access to instances of the Context Broker secured with a PEP Proxy.
+     * For the authentication mechanism to work, the authentication attribute in the configuration has to be fully
+     * configured, and the authentication.enabled subattribute should have the value `true`.
+     *
+     * The Username and password should be considered as sensitive data and should not be stored in plaintext. 
+     * Either encrypt the config and decrypt when initializing the instance or use environment variables secured by
+     * docker secrets.
+     */
+    //authentication: {
+        //enabled: false,
+        /**
+        * Type of the Identity Manager which is used when authenticating the IoT Agent.
+        */
+        //type: 'keystone',
+        /**
+        * Name of the additional header passed to hold the identity of the IoT Agent
+        */
+        //header: 'X-Auth-Token',
+        /**
+        * Hostname of the Identity Manager.
+        */
+        //host: 'localhost',
+        /**
+        * Port of the Identity Manager.
+        */
+        //port: '5000',
+        /**
+        * Username for the IoT Agent - Note this should not be stored in plaintext.
+        */
+        //user: 'IOTA_AUTH_USER',
+        /**
+        * Password for the IoT Agent - Note this should not be stored in plaintext.
+        */
+        //password: 'IOTA_AUTH_PASSWORD'
+    //},
+   
     /**
      * Defines the configuration for the Device Registry, where all the information about devices and configuration
      * groups will be stored. There are currently just two types of registries allowed:
--- a/config.js
+++ b/config.js
@@ -87,6 +87,20 @@ config.amqp = {
     options: {durable: true}
 };
 
+/**
+ * Configuration for the HTTP transport binding.
+ */
+config.http = {
+    /**
+     * South Port where the Ultralight transport binding for HTTP will be listening for device requests.
+     */
+    port: 7896,
+    /**
+     * HTTP Timeout for the http command endpoint (in miliseconds).
+     */
+    //timeout: 1000
+};
+
 config.iota = {
     /**
      * Configures the log level. Appropriate values are: FATAL, ERROR, INFO, WARN and DEBUG.
--- a/config.js
+++ b/config.js
@@ -120,6 +120,15 @@ config.iota = {
      */
         port: 4041
     },
+    authentication: {
+        enabled: true,
+        type: 'keystone',
+        header: 'X-Auth-Token',
+        host: 'localhost',
+        port: '5000',
+        user: 'iotagent',
+        password: 'iotagent'
+    },
     /**
      * Defines the configuration for the Device Registry, where all the information about devices and configuration
      * groups will be stored. There are currently just two types of registries allowed:
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -27,8 +27,6 @@ MAINTAINER FIWARE IoTAgent Team. Telefónica I+D
 # ENV IOTA_AUTH_USER "iotagent"
 # ENV IOTA_AUTH_PASSWORD "iotagent"
 # ENV IOTA_NORTH_PORT "4041"
-# ENV IOTA_REGISTRY_TYPE "mongodb"
-# ENV IOTA_LOG_LEVEL "debug"
 # ENV IOTA_PROVIDER_URL "http://iot-agent:4041"
 
 # The following RUN command retrieves the source code from GitHub.
@@ -44,7 +42,7 @@ MAINTAINER FIWARE IoTAgent Team. Telefónica I+D
 # replace the whole RUN statement by the following COPY statement in your local source using :
 #
 # COPY . /opt/iotajson/
-# COPY ./entrypoint.sh/
+# COPY ./docker/entrypoint.sh/
 #
 RUN if [ "${DOWNLOAD}" = "latest" ] ; \
 	then \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -56,6 +56,9 @@ RUN if [ "${DOWNLOAD}" = "latest" ] ; \
 
 WORKDIR /opt/iotajson
 
+COPY config.js /
+COPY entrypoint.sh /
+
 RUN \
 	# Ensure that Git is installed prior to running npm install
 	apt-get update && \
@@ -71,5 +74,5 @@ RUN \
 USER node
 ENV NODE_ENV=production
 
-ENTRYPOINT ["pm2-runtime", "bin/iotagent-json"]
+ENTRYPOINT ["pm2-runtime", "bin/iotagent-json", "/entrypoint.sh"]
 CMD ["-- ", "config.js"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,7 +33,6 @@ MAINTAINER FIWARE IoTAgent Team. Telefónica I+D
 # replace the whole RUN statement by the following COPY statement in your local source using :
 #
 # COPY . /opt/iotajson/
-# COPY . /docker/entrypoint.sh/
 #
 RUN if [ "${DOWNLOAD}" = "latest" ] ; \
 	then \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,7 +33,7 @@ MAINTAINER FIWARE IoTAgent Team. Telefónica I+D
 # replace the whole RUN statement by the following COPY statement in your local source using :
 #
 # COPY . /opt/iotajson/
-# COPY ./docker/entrypoint.sh/
+# COPY . /docker/entrypoint.sh/
 #
 RUN if [ "${DOWNLOAD}" = "latest" ] ; \
 	then \
@@ -77,5 +77,5 @@ RUN \
 USER node
 ENV NODE_ENV=production
 
-ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT ["docker/entrypoint.sh"]
 CMD ["-- ", "config.js"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -57,8 +57,8 @@ RUN if [ "${DOWNLOAD}" = "latest" ] ; \
 	apt-get clean && \
 	apt-get remove -y unzip && \
 	apt-get -y autoremove
-
-
+	
+	
 WORKDIR /opt/iotajson
 
 RUN \
@@ -71,7 +71,8 @@ RUN \
 	# Remove Git and clean apt cache
 	apt-get clean && \
 	apt-get remove -y git && \
-	apt-get -y autoremove
+	apt-get -y autoremove && \ 
+	chmod +x docker/entrypoint.sh
 
 USER node
 ENV NODE_ENV=production
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,17 +17,8 @@ MAINTAINER FIWARE IoTAgent Team. Telefónica I+D
 # IMPORTANT: For a Production Environment Use Docker Secrets to define
 #  these values and add _FILE to the name of the variable.
 
-# ENV IOTA_CB_HOST "orion"
-# ENV IOTA_CB_PORT  "1026"
-# ENV IOTA_MONGO_HOST "mongo-db"
-# ENV IOTA_MONGO_PORT  "27017"
-# ENV IOTA_AUTH_ENABLED false
-# ENV IOTA_AUTH_HOST "localhost"
-# ENV IOTA_AUTH_PORT "5000"
 # ENV IOTA_AUTH_USER "iotagent"
 # ENV IOTA_AUTH_PASSWORD "iotagent"
-# ENV IOTA_NORTH_PORT "4041"
-# ENV IOTA_PROVIDER_URL "http://iot-agent:4041"
 
 # The following RUN command retrieves the source code from GitHub.
 # 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,7 +14,23 @@ ENV DOWNLOAD=${DOWNLOAD}
 
 MAINTAINER FIWARE IoTAgent Team. Telefónica I+D
 
-#
+# IMPORTANT: For a Production Environment Use Docker Secrets to define
+#  these values and add _FILE to the name of the variable.
+
+# ENV IOTA_CB_HOST "orion"
+# ENV IOTA_CB_PORT  "1026"
+# ENV IOTA_MONGO_HOST "mongo-db"
+# ENV IOTA_MONGO_PORT  "27017"
+# ENV IOTA_AUTH_ENABLED false
+# ENV IOTA_AUTH_HOST "localhost"
+# ENV IOTA_AUTH_PORT "5000"
+# ENV IOTA_AUTH_USER "iotagent"
+# ENV IOTA_AUTH_PASSWORD "iotagent"
+# ENV IOTA_NORTH_PORT "4041"
+# ENV IOTA_REGISTRY_TYPE "mongodb"
+# ENV IOTA_LOG_LEVEL "debug"
+# ENV IOTA_PROVIDER_URL "http://iot-agent:4041"
+
 # The following RUN command retrieves the source code from GitHub.
 # 
 # To obtain the latest stable release run this Docker file with the parameters
@@ -28,6 +44,7 @@ MAINTAINER FIWARE IoTAgent Team. Telefónica I+D
 # replace the whole RUN statement by the following COPY statement in your local source using :
 #
 # COPY . /opt/iotajson/
+# COPY ./entrypoint.sh/
 #
 RUN if [ "${DOWNLOAD}" = "latest" ] ; \
 	then \
@@ -56,9 +73,6 @@ RUN if [ "${DOWNLOAD}" = "latest" ] ; \
 
 WORKDIR /opt/iotajson
 
-COPY config.js /
-COPY entrypoint.sh /
-
 RUN \
 	# Ensure that Git is installed prior to running npm install
 	apt-get update && \
@@ -74,5 +88,5 @@ RUN \
 USER node
 ENV NODE_ENV=production
 
-ENTRYPOINT ["pm2-runtime", "bin/iotagent-json", "/entrypoint.sh"]
+ENTRYPOINT ["/entrypoint.sh"]
 CMD ["-- ", "config.js"]
--- a/docker/README.md
+++ b/docker/README.md
@@ -150,3 +150,12 @@ COPY . /opt/iotajson/
 ```
 
 Full instructions can be found within the `Dockerfile` itself.
+
+### Docker Secrets
+
+As an alternative to passing sensitive information via environment variables, `_FILE` may be appended to some sensitive environment variables, causing the initialization script to load the values for those variables from files present in the container. In particular, this can be used to load passwords from Docker secrets stored in `/run/secrets/<secret_name>` files.
+
+Currently, this the `_FILE` suffix is supported for:
+
+* `IOTA_AUTH_USER`
+* `IOTA_AUTH_PASSWORD`
--- a/docker/README.md
+++ b/docker/README.md
@@ -156,7 +156,6 @@ Full instructions can be found within the `Dockerfile` itself.
 As an alternative to passing sensitive information via environment variables, `_FILE` may be appended to some sensitive environment variables, causing the initialization script to load the values for those variables from files present in the container. In particular, this can be used to load passwords from Docker secrets stored in `/run/secrets/<secret_name>` files.
 For example:
 
-Currently, this the `_FILE` suffix is supported for:
 ```console
 docker run --name iotagent -e IOTA_AUTH_PASSWORD_FILE=/run/secrets/password -d fiware/iotagent-json
 ```
--- a/docker/README.md
+++ b/docker/README.md
@@ -154,6 +154,12 @@ Full instructions can be found within the `Dockerfile` itself.
 ### Docker Secrets
 
 As an alternative to passing sensitive information via environment variables, `_FILE` may be appended to some sensitive environment variables, causing the initialization script to load the values for those variables from files present in the container. In particular, this can be used to load passwords from Docker secrets stored in `/run/secrets/<secret_name>` files.
+For example:
+
+Currently, this the `_FILE` suffix is supported for:
+```console
+docker run --name iotagent -e IOTA_AUTH_PASSWORD_FILE=/run/secrets/password -d fiware/iotagent-json
+```
 
 Currently, this the `_FILE` suffix is supported for:
 
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -30,7 +30,7 @@ if [[  -z "$IOTA_AUTH_ENABLED" ]]; then
  echo "WARNING: It is recommended to enable authentication for secure connection"
  echo "***********************************************"
 else
-    if [[ -z "$IOTA_AUTH_PASSWORD" ]]; then
+    if  [[ -z "$IOTA_AUTH_USER" ]] || [ -z "$IOTA_AUTH_PASSWORD" ]]; then
         echo "***********************************************"
         echo "WARNING: Default IoT Agent Auth credentials have not been overridden"
         echo "The connection to the Context broker fails"
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -62,4 +62,4 @@ else
     fi
 fi
 
-pm2-runtime start /bin/iotagent-json
+pm2-runtime opt/iotajson/bin/iotagent-json
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,4 +1,25 @@
 #!/bin/bash
+#
+# Copyright 2019 Telefonica Investigación y Desarrollo, S.A.U
+#
+# This file is part of iotagent-json
+#
+# iotagent-json is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the License,
+# or (at your option) any later version.
+#
+# iotagent-json is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public
+# License along with iotagent-ul. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by the GNU Affero General Public License please contact
+# with iot_support at tid dot es
+#
 
 # usage: file_env VAR [DEFAULT]
 #    ie: file_env 'XYZ_DB_PASSWORD' 'example'
@@ -33,7 +54,6 @@ else
     if  [[ -z "$IOTA_AUTH_USER" ]] || [ -z "$IOTA_AUTH_PASSWORD" ]]; then
         echo "***********************************************"
         echo "WARNING: Default IoT Agent Auth credentials have not been overridden"
-        echo "The connection to the Context broker fails"
         echo "***********************************************"
     else
         echo "***********************************************"
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -62,4 +62,4 @@ else
     fi
 fi
 
-pm2-runtime opt/iotajson/bin/iotagent-json
+pm2-runtime bin/iotagent-json
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -22,21 +22,18 @@ file_env() {
     unset "$fileVar"
 }
 
-file_env 'IOTA_CB_HOST'
-file_env 'IOTA_CB_PORT'
-file_env 'IOTA_MONGO_HOST'
-file_env 'IOTA_MONGO_PORT'
 file_env 'IOTA_AUTH_USER'
 file_env 'IOTA_AUTH_PASSWORD'
 
-if [[ $IOTA_AUTH_USER ==  "iotagent" ]] || [[ $IOTA_AUTH_PASSWORD == "iotagent'" ]]; then
+if [[ $IOTA_AUTH_USER ==  config.getconfig().iota.authentication.user ]] || [[ $IOTA_AUTH_PASSWORD == config.getconfig().iota.authentication.password ]]; then
         echo "***********************************************"
-        echo "WARNING: It is recommended to change IOTA Auth credentials not to use default values"
+        echo "WARNING: It is recommended to set IOTA Auth credentials(using difficult pattern) keeping security in mind and not to use default values"
         echo "These keys should be set using Docker Secrets"
         echo "***********************************************"
 fi
 
-#cat /bin/iotagent-json
+node pm2-runtime
+node /bin/iotagent-json
 
 exec /sbin/init
 
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -25,17 +25,19 @@ file_env() {
 file_env 'IOTA_AUTH_USER'
 file_env 'IOTA_AUTH_PASSWORD'
 
-if [[ $IOTA_AUTH_USER ==  config.getconfig().iota.authentication.user ]] || [[ $IOTA_AUTH_PASSWORD == config.getconfig().iota.authentication.password ]]; then
-        echo "***********************************************"
-        echo "WARNING: It is recommended to set IOTA Auth credentials(using difficult pattern) keeping security in mind and not to use default values"
-        echo "These keys should be set using Docker Secrets"
-        echo "***********************************************"
-fi
+# Reading username and password at runtime
+read -p "Username : " IOTA_AUTH_USER
+read -p "Password : " IOTA_AUTH_PASSWORD
+echo "***********************************************"
+echo "WARNING: It is recommended to set IOTA Auth credentials(using difficult pattern) keeping security in mind and not to use default values"
+echo "These keys should be set using Docker Secrets"
+echo "***********************************************"
 
-node pm2-runtime
-node /bin/iotagent-json
+#Setting values to ENV vars IOTA_AUTH_PASSWORD & IOTA_AUTH_USER
+export IOTA_AUTH_PASSWORD=$IOTA_AUTH_PASSWORD
+export IOTA_AUTH_USER=$IOTA_AUTH_USER
 
-exec /sbin/init
-
+npm install -g pm2
+pm2 start /bin/iotagent-json
 
 #/bin/bash
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -25,19 +25,16 @@ file_env() {
 file_env 'IOTA_AUTH_USER'
 file_env 'IOTA_AUTH_PASSWORD'
 
-# Reading username and password at runtime
-read -p "Username : " IOTA_AUTH_USER
-read -p "Password : " IOTA_AUTH_PASSWORD
-echo "***********************************************"
-echo "WARNING: It is recommended to set IOTA Auth credentials(using difficult pattern) keeping security in mind and not to use default values"
-echo "These keys should be set using Docker Secrets"
-echo "***********************************************"
-
-#Setting values to ENV vars IOTA_AUTH_PASSWORD & IOTA_AUTH_USER
 export IOTA_AUTH_PASSWORD=$IOTA_AUTH_PASSWORD
 export IOTA_AUTH_USER=$IOTA_AUTH_USER
 
-npm install -g pm2
+if [[ $IOTA_AUTH_USER ==  "" ]] || [[ $IOTA_AUTH_PASSWORD == "" ]]; then
+        echo "***********************************************"
+        echo "WARNING: It is recommended to set IOTA Auth credentials(using difficult pattern) keeping security in mind and not to use default values"
+        echo "These keys should be set using Docker Secrets"
+        echo "***********************************************"
+fi
+
 pm2 start /bin/iotagent-json
 
-#/bin/bash
+/bin/bash
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -36,6 +36,11 @@ else
         echo "WARNING: It is recommended to set IOTA Auth credentials when using authentication"
         echo "These keys should be set using Docker Secrets"
         echo "***********************************************"
+    else
+        echo "***********************************************"
+        echo "WARNING: It is recommended to use difficult passwords keeping security in mind and not use the default ones"
+        echo "These keys should be set using Docker Secrets"
+        echo "***********************************************"
     fi
 fi
 
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,0 +1,61 @@
+#!/bin/sh
+
+file_env() {
+    local var="$1"
+    local fileVar="${var}_FILE"
+    local def="${2:-}"
+    if [ "${!var:-}" ] && [ "${!fileVar:-}" ]; then
+        echo >&2 "error: both $var and $fileVar are set (but are exclusive)"
+        exit 1
+    fi
+    local val="$def"
+    if [ "${!var:-}" ]; then
+        val="${!var}"
+    elif [ "${!fileVar:-}" ]; then
+        val="$(< "${!fileVar}")"
+    fi
+    export "$var"="$val"
+    unset "$fileVar"
+}
+
+if [[ ${IOTA_MONGO_PORT} =~ ^http://(.*):(.*)+$ ]] ; then
+    export IOTA_MONGO_PORT=${BASH_REMATCH[2]}
+    export IOTA_MONGO_HOST=${BASH_REMATCH[1]}
+    
+fi
+
+if [[ ${IOTA_CB_PORT} =~ ^http://(.*):(.*)+$ ]] ; then
+    export IOTA_CB_PORT=${BASH_REMATCH[2]}
+    export IOTA_CB_HOST=${BASH_REMATCH[1]}
+fi
+
+export HOST_IP=`awk 'NR==1{print $1}' /etc/hosts`
+
+echo "HOST IP: $HOST_IP"
+echo "MONGODB HOST: $IOTA_MONGO_HOST"
+echo "MONGODB PORT: $IOTA_MONGO_PORT"
+echo "ORION CB HOST: $IOTA_CB_HOST"
+echo "ORION CB PORT: $IOTA_CB_PORT"
+
+
+sed -i /lib/configService.js \
+        -e "s|IOTA_CB_HOST|${IOTA_CB_HOST}|g" \
+        -e "s|IOTA_CB_PORT|${IOTA_CB_PORT}|g" \
+        -e "s|IOTA_MONGO_HOST|${IOTA_MONGO_HOST}|g" \
+    -e "s|IOTA_MONGO_PORT|${IOTA_MONGO_PORT}|g" \
+    -e "s|IOTA_AUTH_HOST|${IOTA_AUTH_HOST}|g" \
+    -e "s|IOTA_AUTH_PORT|${IOTA_AUTH_PORT}|g" \
+    -e "s|IOTA_AUTH_USER|${IOTA_AUTH_USER}|g" \
+    -e "s|IOTA_AUTH_PASSWORD|${IOTA_AUTH_PASSWORD}|g" \
+    -e "s|IOTA_AUTH_PERMANENT_TOKEN|${IOTA_AUTH_PERMANENT_TOKEN}|g" \
+    -e "s|IOTA_PROVIDER_URL|${IOTA_PROVIDER_URL}|g" \
+    -e "s|IOTA_IOTAM_HOST|${IOTA_IOTAM_HOST}|g"
+
+
+
+exec /sbin/init
+
+
+
+
+#/bin/bash
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -42,4 +42,4 @@ else
     fi
 fi
 
-/bin/bash
+pm2 start /bin/iotagent-json
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,5 +1,9 @@
 #!/bin/sh
 
+# usage: file_env VAR [DEFAULT]
+#    ie: file_env 'XYZ_DB_PASSWORD' 'example'
+# (will allow for "$XYZ_DB_PASSWORD_FILE" to fill in the value of
+#  "$XYZ_DB_PASSWORD" from a file, especially for Docker's secrets feature)
 file_env() {
     local var="$1"
     local fileVar="${var}_FILE"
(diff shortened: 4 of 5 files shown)


Jagatjot:
@fgalan @jason-fox can you please rewiew the PR ?

lib/configService.js
```diff
@@ -55,7 +56,21 @@ function processEnvironmentVariables() {
             'IOTA_AMQP_RETRY_TIME',
             'IOTA_HTTP_HOST',
             'IOTA_HTTP_PORT',
-            'IOTA_HTTP_TIMEOUT'
+            'IOTA_HTTP_TIMEOUT',
+            'IOTA_CB_URL',
```
fgalan: I'm not fully sure, but I'd say that all these env var are the ones that are actually managed by IOTA Node Lib (for instance, see https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/commonConfig.js#L115). IOTA specific code shouldn't care about them.

docker/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/sh
```
fgalan: As far as I understand this script is a kind of translator between config.js and env vars to be provided to the agent when it runs inside docker container.

Why don't just inject the env vars directly (i.e. ENV)? I don't see the point in this extra complexity... but maybe I'm missing something :)

Jagatjot: Yes your understanding is right.

For script I followed the same pattern as suggested in Issue #366 and https://github.com/telefonicaid/iotagent-node-lib/pull/726#discussion_r242214046 for Cygnus and PostGres.

lib/configService.js
Jagatjot: I think you are right so should I remove the changes done in this file ?

fgalan: Yes, I think so.

Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/f78515c77f7455bbde2fce8b806b85c99ecda014

config.js
```diff
@@ -134,6 +134,15 @@ config.iota = {
      */
         port: 4041
     },
+    authentication: {
```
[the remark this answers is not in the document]
Jagatjot: As authentication feature has been implemented for IOTAs so I thought of adding this to default config. May be we could keep default enabled to false or keep it commented ?

config.js
```diff
@@ -134,6 +134,15 @@ config.iota = {
...
+        enabled: true,
+        type: 'keystone',
```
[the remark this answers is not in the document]
Jagatjot: The official documentation of iotagent-node-lib says it is keystone https://github.com/telefonicaid/iotagent-node-lib/blob/4aae8dd9ccac738569a10bffedd4b8567dc251f7/doc/installationguide.md

config.js
```diff
@@ -134,6 +134,15 @@ config.iota = {
...
+        header: 'X-Auth-Token',
+        host: 'localhost',
+        port: '5000',
+        user: 'iotagent',
```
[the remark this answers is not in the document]
Jagatjot: In official documentation of iotagent-node-lib , they are passed as above https://github.com/telefonicaid/iotagent-node-lib/blob/4aae8dd9ccac738569a10bffedd4b8567dc251f7/doc/installationguide.md . Should I change it to plain text ?

config.js
```diff
@@ -134,6 +134,15 @@ config.iota = {
      */
         port: 4041
     },
+    authentication: {
+        enabled: true,
+        type: 'keystone',
+        header: 'X-Auth-Token',
+        host: 'localhost',
```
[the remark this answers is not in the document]
Jagatjot: In official documentation of iotagent-node-lib, the default url is passed as localhost (https://github.com/telefonicaid/iotagent-node-lib/blob/4aae8dd9ccac738569a10bffedd4b8567dc251f7/doc/installationguide.md)

docker/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+export HOST_IP=`awk 'NR==1{print $1}' /etc/hosts`
+
+echo "HOST IP: $HOST_IP"
+echo "MONGODB HOST: $IOTA_MONGO_HOST"
+echo "MONGODB PORT: $IOTA_MONGO_PORT"
+echo "ORION CB HOST: $IOTA_CB_HOST"
+echo "ORION CB PORT: $IOTA_CB_PORT"
+
+
+sed -i /lib/configService.js \
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/5997db35549eb78b42343cdbe8c572ed775c7645

docker/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/sh
+
+file_env() {
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/5997db35549eb78b42343cdbe8c572ed775c7645

docker/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+    local val="$def"
+    if [ "${!var:-}" ]; then
+        val="${!var}"
+    elif [ "${!fileVar:-}" ]; then
+        val="$(< "${!fileVar}")"
+    fi
+    export "$var"="$val"
+    unset "$fileVar"
+}
+
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/5997db35549eb78b42343cdbe8c572ed775c7645

lib/configService.js
```diff
@@ -26,6 +26,7 @@
 var config = {},
     logger = require('logops');
 
+
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/e0a18a6a89feaeb85712bc88a680b0ae18ef4bda

lib/configService.js
```diff
@@ -82,6 +83,7 @@ function processEnvironmentVariables() {
             'IOTA_HTTP_TIMEOUT'
         ];
 
+
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/e0a18a6a89feaeb85712bc88a680b0ae18ef4bda

docker/Dockerfile
```diff
@@ -71,5 +74,5 @@ RUN \
 USER node
 ENV NODE_ENV=production
 
-ENTRYPOINT ["pm2-runtime", "bin/iotagent-json"]
+ENTRYPOINT ["pm2-runtime", "bin/iotagent-json", "/entrypoint.sh"]
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/f4805374ae605b0c21371fdf66a8320533d198b9

docker/Dockerfile
```diff
@@ -56,6 +56,9 @@ RUN if [ "${DOWNLOAD}" = "latest" ] ; \
 
 WORKDIR /opt/iotajson
 
+COPY config.js /
+COPY entrypoint.sh /
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/f4805374ae605b0c21371fdf66a8320533d198b9

docker/Dockerfile
```diff
@@ -56,6 +56,9 @@ RUN if [ "${DOWNLOAD}" = "latest" ] ; \
 
 WORKDIR /opt/iotajson
 
+COPY config.js /
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/f4805374ae605b0c21371fdf66a8320533d198b9

docker/entrypoint.sh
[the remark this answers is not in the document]
Jagatjot: I have removed the overcomplications from the script. Can you please review and suggest any further changes required. Thank you!

docker/entrypoint.sh
```diff
@@ -0,0 +1,44 @@
...
+file_env 'IOTA_CB_HOST'
+file_env 'IOTA_CB_PORT'
+file_env 'IOTA_MONGO_HOST'
+file_env 'IOTA_MONGO_PORT'
+file_env 'IOTA_AUTH_USER'
+file_env 'IOTA_AUTH_PASSWORD'
+
+if [[ $IOTA_AUTH_USER ==  "iotagent" ]] || [[ $IOTA_AUTH_PASSWORD == "iotagent'" ]]; then
```
jason-fox: The `echo` here should do **one** of the following:

1. If we're running without protection echo a warning.
2. If we're running with protection but haven't overridden the conf ENV variables, echo a warning (it will  probably fail to connect)
3. If we're running with protection, have overridden the conf ENV, but have used plain text ENVs, echo a warning

In otherwords encourage good practice, but don't spam the user with irrelevant start up messages.

docker/entrypoint.sh
```diff
@@ -0,0 +1,44 @@
...
+        echo "***********************************************"
+        echo "WARNING: It is recommended to change IOTA Auth credentials not to use default values"
```
jason-fox: This statement is wrong - without an override, the IoT Agent will fail to connect - it should be an ERROR

docker/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+fi
+
+if [[ ${IOTA_CB_PORT} =~ ^http://(.*):(.*)+$ ]] ; then
+    export IOTA_CB_PORT=${BASH_REMATCH[2]}
+    export IOTA_CB_HOST=${BASH_REMATCH[1]}
+fi
+
+export HOST_IP=`awk 'NR==1{print $1}' /etc/hosts`
+
+echo "HOST IP: $HOST_IP"
```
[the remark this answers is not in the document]
jason-fox: LGTM

docker/Dockerfile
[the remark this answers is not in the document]
jason-fox: LGTM

docker/entrypoint.sh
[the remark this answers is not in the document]
jason-fox: LGTM

docker/entrypoint.sh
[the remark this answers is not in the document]
jason-fox: LGTM , though I'm not sure if `IOTA_CB_HOST, IOTA_CB_PORT, IOTA_MONGO_HOST` or `IOTA_MONGO_PORT` are really necessary

lib/configService.js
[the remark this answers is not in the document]
jason-fox: LGTM

lib/configService.js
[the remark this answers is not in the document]
jason-fox: LGTM

docker/Dockerfile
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/df770138c08a0de85d4a79a673af0d1477ec4d8e

config.js
```diff
@@ -134,6 +134,15 @@ config.iota = {
      */
         port: 4041
     },
+    authentication: {
+        enabled: true,
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/356f621cb520a5007f219beb0316e5e12dae28cc

docker/Dockerfile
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/e0d43d1f1a95753e61b5d4b89edd118c1282a20b

config.js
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/356f621cb520a5007f219beb0316e5e12dae28cc

docker/entrypoint.sh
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/df770138c08a0de85d4a79a673af0d1477ec4d8e

docker/Dockerfile
```diff
@@ -71,5 +88,5 @@ RUN \
 USER node
 ENV NODE_ENV=production
 
-ENTRYPOINT ["pm2-runtime", "bin/iotagent-json"]
+ENTRYPOINT ["/entrypoint.sh"]
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/df770138c08a0de85d4a79a673af0d1477ec4d8e

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/df770138c08a0de85d4a79a673af0d1477ec4d8e

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/df770138c08a0de85d4a79a673af0d1477ec4d8e

docker/Dockerfile
```diff
@@ -14,7 +14,23 @@ ENV DOWNLOAD=${DOWNLOAD}
 
 MAINTAINER FIWARE IoTAgent Team. Telefónica I+D
 
-#
+# IMPORTANT: For a Production Environment Use Docker Secrets to define
+#  these values and add _FILE to the name of the variable.
+
+# ENV IOTA_CB_HOST "orion"
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/e0d43d1f1a95753e61b5d4b89edd118c1282a20b

docker/Dockerfile
```diff
@@ -28,6 +44,7 @@ MAINTAINER FIWARE IoTAgent Team. Telefónica I+D
 # replace the whole RUN statement by the following COPY statement in your local source using :
 #
 # COPY . /opt/iotajson/
+# COPY ./entrypoint.sh/
```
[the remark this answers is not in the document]
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/e0d43d1f1a95753e61b5d4b89edd118c1282a20b

Jagatjot:
@jason-fox Done the changes asked by you. Any further suggestions? Thank You!

config.js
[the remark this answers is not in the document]
jason-fox:
```
// Usename and password must be overriden by setting them using difficult pattern for security issues.
```

There are two typos in this, and it doesn't clearly describe the fundamental reasons for changing them.

1. The  Username and Password will be supplied by your IDM, please replace the dummy values below
2.  Do not store Usernames and Passwords in plain text on a production system - it is insecure
3. Either overwrite the config at runtime or User Docker Secrets to protect your username and password.

docker/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+    -e "s|IOTA_AUTH_PORT|${IOTA_AUTH_PORT}|g" \
+    -e "s|IOTA_AUTH_USER|${IOTA_AUTH_USER}|g" \
+    -e "s|IOTA_AUTH_PASSWORD|${IOTA_AUTH_PASSWORD}|g" \
+    -e "s|IOTA_AUTH_PERMANENT_TOKEN|${IOTA_AUTH_PERMANENT_TOKEN}|g" \
+    -e "s|IOTA_PROVIDER_URL|${IOTA_PROVIDER_URL}|g" \
+    -e "s|IOTA_IOTAM_HOST|${IOTA_IOTAM_HOST}|g"
+
+
+
+exec /sbin/init
```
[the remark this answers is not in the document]
jason-fox: https://pm2.io/runtime/ uses the following:

```bash
npm install -g pm2
pm2 start app.js  # I assume in our case  /bin/iotagent-json
```

I'm not sure what the following is meant to be doing.

```
node pm2-runtime
node /bin/iotagent-json
exec /sbin/init
```

How have you tested this?

docker/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+        echo "***********************************************"
+fi
+
+node pm2-runtime
+node /bin/iotagent-json
+
+exec /sbin/init
+
+
+#/bin/bash
```
jason-fox: Remove blank lines and comment

lib/configService.js
```diff
@@ -193,7 +193,9 @@ function getLogger() {
     return logger;
 }
 
+
```
jason-fox: Extraneous Whitespace

lib/configService.js
```diff
@@ -193,7 +193,9 @@ function getLogger() {
...
 exports.setConfig = setConfig;
 exports.getConfig = getConfig;
 exports.setLogger = setLogger;
 exports.getLogger = getLogger;
+
```
jason-fox: Extraneous Whitespace

docker/Dockerfile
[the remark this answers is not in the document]
jason-fox: This change is fixed so LGTM, however there are still further issues in `entrypoint.sh`

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/aa92736eb457d1d41fb476e856569020924c7ba6

lib/configService.js
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/6c90fc5c357f9c8bc6b6d1df892d6118cb7043b4

lib/configService.js
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/6c90fc5c357f9c8bc6b6d1df892d6118cb7043b4

Jagatjot:
@jason-fox I have updated the entrypoint.sh file to read username and password at runtime which will override the current default values and export those values as Env variables. Thank You!

docker/entrypoint.sh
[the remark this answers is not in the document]
jason-fox: LGTM - better to keep the change minimal

docker/entrypoint.sh
```diff
@@ -0,0 +1,43 @@
...
+echo "***********************************************"
+echo "WARNING: It is recommended to set IOTA Auth credentials(using difficult pattern) keeping security in mind and not to use default values"
+echo "These keys should be set using Docker Secrets"
+echo "***********************************************"
+
+#Setting values to ENV vars IOTA_AUTH_PASSWORD & IOTA_AUTH_USER
+export IOTA_AUTH_PASSWORD=$IOTA_AUTH_PASSWORD
+export IOTA_AUTH_USER=$IOTA_AUTH_USER
+
+npm install -g pm2
```
jason-fox: Right line, wrong place. This already exists in the `Dockerfile` see https://github.com/telefonicaid/iotagent-json/blob/346886/docker/Dockerfile#L69

Please delete (as git will have been removed at this point in the operation

docker/entrypoint.sh
jason-fox: The comment `#/bin/bash` still needs to be removed

docker/entrypoint.sh
```diff
@@ -0,0 +1,43 @@
...
+        val="$(< "${!fileVar}")"
+    fi
+    export "$var"="$val"
+    unset "$fileVar"
+}
+
+file_env 'IOTA_AUTH_USER'
+file_env 'IOTA_AUTH_PASSWORD'
+
+# Reading username and password at runtime
```
jason-fox: @Jagatjot  - I forgot this one:

This section is incorrect and needs amendment. You don't need to read Username and password here and it doesn't resolve the previous comments.

1. If you are not running in secure mode, warn
2. If you are running in secure mode and haven't added Username/Password warn

The text about a difficult pattern is incorrect. The Username/Password will be defined for you.

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/151094fb44fab57ec2ba01e6b9c170c31ceb3c42

docker/entrypoint.sh
Jagatjot: I have kept username/password to null in this https://github.com/telefonicaid/iotagent-json/pull/380/commits/151094fb44fab57ec2ba01e6b9c170c31ceb3c42 and printed a warning to set the auth credentials?

jason-fox: The text 

> It is recommended to set IOTA Auth credentials(using difficult pattern) 

Still doesn't make sense - The Username/Password will be defined for you.

These lines:

```bash
export IOTA_AUTH_PASSWORD=$IOTA_AUTH_PASSWORD
export IOTA_AUTH_USER=$IOTA_AUTH_USER
```

repeat the work of the  `fileEnv()` function  - delete


This if-then

```bash
if [[ $IOTA_AUTH_USER ==  "" ]] || [[ $IOTA_AUTH_PASSWORD == "" ]]; then
```

needs to be an if-then-else - see https://github.com/telefonicaid/iotagent-json/pull/380#discussion_r254205868

* First check if the Auth is enabled (and if not echo a warning)
* Then for users in secure mode only check if  IOTA_AUTH_USER + IOTA_AUTH_PASSWORD are set (not blank) and echo the other warning.

If I am not running in secure mode I don;t want an irrelevant warning about setting a username I don't need.

jason-fox: > @jason-fox from where the username/password be coming in entrypoint.sh file. Will it come from config.js file if we do not read them here ?

They are supplied as  `ENV` variables when instantiating the docker container - see [this example](https://github.com/telefonicaid/iotagent-json/tree/master/docker). Or indeed as **Docker Secrets** - encrypted files with an associated `ENV` variable - in other words the whole idea of this issue - avoiding placing plain text sensitive data into a file (or an `ENV` which can then be inspected.

If you use the **Docker Secrets** method, all you get is the file path to an encrypted file.

If you don't use `ENV` variables then the `config.js` values would be used as default You could of course use your own `config.js` (by attaching it as a `VOLUME`) - but once again you'll end up with the passwords in plain text within a file.

Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/efc36a520e965bd17fe0b21d9ac7bb202431504c

Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/c6aa3ab7d521f562f849e863dc83272cffdbce37

Jagatjot: @jason-fox I have updated the changes in https://github.com/telefonicaid/iotagent-json/pull/380/commits/c6aa3ab7d521f562f849e863dc83272cffdbce37 . It first check if auth is enabled or not(if not, its gives a warning to use authentication), if it is enabled but the username and password are not set(it gives the warning to set username and password) and if username/password are set, it gives the warning to use difficult passwords. Thank you!
I

docker/entrypoint.sh
```diff
@@ -0,0 +1,49 @@
...
+    fi
+    export "$var"="$val"
+    unset "$fileVar"
+}
+
+file_env 'IOTA_AUTH_USER'
+file_env 'IOTA_AUTH_PASSWORD'
+file_env 'IOTA_AUTH_ENABLED'
+
+if [[ $IOTA_AUTH_ENABLED ==  "false" ]]; then
```
jason-fox: hmmm... this still doesn't quite make sense, you're saying if I **deliberately set** the `ENV` to `false` then the component should nag you, but `false` is the default anyway so why would anyone set it?

It should be a empty check not an equals check - If I deliberately override the `ENV` to switch off security I don't need to be nagged, if I **forget** to switch it on I do.

docker/entrypoint.sh
```diff
@@ -0,0 +1,49 @@
...
+ echo "***********************************************"
+ echo "WARNING: It is recommended to enable authentication for secure connection"
+ echo "***********************************************"
+else
+    if [[ $IOTA_AUTH_USER ==  "" ]] || [[ $IOTA_AUTH_PASSWORD == "" ]]; then
```
jason-fox: Once again the comparison is incorrect - it should be an empty check, not a string comparison looking for a blank

docker/entrypoint.sh
```diff
@@ -0,0 +1,49 @@
...
+        echo "***********************************************"
+        echo "WARNING: It is recommended to set IOTA Auth credentials when using authentication"
+        echo "These keys should be set using Docker Secrets"
```
jason-fox: By this point you don't know if you're dealing with `IOTA_AUTH_USER` or `IOTA_AUTH_USER_FILE` - The IOTA_AUTH_USER will have been set earlier in the script. Delete this line

docker/entrypoint.sh
```diff
@@ -0,0 +1,49 @@
...
+file_env 'IOTA_AUTH_ENABLED'
+
+if [[ $IOTA_AUTH_ENABLED ==  "false" ]]; then
+ echo "***********************************************"
+ echo "WARNING: It is recommended to enable authentication for secure connection"
+ echo "***********************************************"
+else
+    if [[ $IOTA_AUTH_USER ==  "" ]] || [[ $IOTA_AUTH_PASSWORD == "" ]]; then
+        echo "***********************************************"
+        echo "WARNING: It is recommended to set IOTA Auth credentials when using authentication"
```
jason-fox: The Warning should read:

> WARNING: Default IoT Agent Auth credentials have not been overridden

And nothing more - we can't infer too much from this.
The connection to the Context broker will (probably) **Fail** directly after this.

Docker is flexible so I know there is a workaround where you could supply a `VOLUME`, but its more likely that you've set Auth on and not supplied the Username/Password

docker/entrypoint.sh
```diff
@@ -0,0 +1,49 @@
...
+        echo "These keys should be set using Docker Secrets"
+        echo "***********************************************"
+    else
+        echo "***********************************************"
+        echo "WARNING: It is recommended to use difficult passwords keeping security in mind and not use the default ones"
+        echo "These keys should be set using Docker Secrets"
```
jason-fox: By this point you don't know if you're dealing with `IOTA_AUTH_USER` or `IOTA_AUTH_USER_FILE` - The IOTA_AUTH_USER will have been set earlier in the script. Delete this line

docker/entrypoint.sh
```diff
@@ -0,0 +1,49 @@
...
+ echo "***********************************************"
+else
+    if [[ $IOTA_AUTH_USER ==  "" ]] || [[ $IOTA_AUTH_PASSWORD == "" ]]; then
+        echo "***********************************************"
+        echo "WARNING: It is recommended to set IOTA Auth credentials when using authentication"
+        echo "These keys should be set using Docker Secrets"
+        echo "***********************************************"
+    else
+        echo "***********************************************"
+        echo "WARNING: It is recommended to use difficult passwords keeping security in mind and not use the default ones"
```
jason-fox: The Warning should read:

> INFO: IoT Agent Auth credentials have been overridden

And nothing more - we can't infer too much from this. Maybe they have proper passwords already?

docker/entrypoint.sh
Jagatjot: Isn't this the way to check if auth is enabled or not as suggested in comment https://github.com/telefonicaid/iotagent-json/pull/380#discussion_r254249337 ?

jason-fox: No - read it carefully - emphasis added:

> * Then for users in secure mode only check if  IOTA_AUTH_USER + IOTA_AUTH_PASSWORD **are set (not blank)** and echo the other warning.

Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/c33b866240d67dcd3071bf277b278e1e617c5c9f

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/c33b866240d67dcd3071bf277b278e1e617c5c9f

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/c33b866240d67dcd3071bf277b278e1e617c5c9f

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/c33b866240d67dcd3071bf277b278e1e617c5c9f

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/c33b866240d67dcd3071bf277b278e1e617c5c9f

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/c33b866240d67dcd3071bf277b278e1e617c5c9f

Jagatjot:
@jason-fox in https://github.com/telefonicaid/iotagent-json/pull/380/commits/c33b866240d67dcd3071bf277b278e1e617c5c9f I have checked if auth is not enbaled then is else if I have checked if password is null then it gives an error and in else statement if password is set it gives INFO: IoT Agent Auth credentials have been overridden.

config.js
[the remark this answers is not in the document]
jason-fox: NTC

config.js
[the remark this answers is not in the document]
jason-fox: NTC - The documentation and `keyrock/keystone` should be raised as a separate issue if necessary

config.js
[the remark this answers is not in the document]
jason-fox: NTC

docker/entrypoint.sh
```diff
@@ -0,0 +1,47 @@
...
+
+file_env 'IOTA_AUTH_USER'
+file_env 'IOTA_AUTH_PASSWORD'
+
+if [[  -z "$IOTA_AUTH_ENABLED" ]]; then
+ echo "***********************************************"
+ echo "WARNING: It is recommended to enable authentication for secure connection"
+ echo "***********************************************"
+else
+    if [[ -z "$IOTA_AUTH_PASSWORD" ]]; then
```
jason-fox: You have removed the check on the Username - please re-instate it

Jagatjot: I guess now it must be LGTM after https://github.com/telefonicaid/iotagent-json/pull/380/commits/0fbd0024ee2c28f66421dea6fc6e88dfd59faa9d 😀

jason-fox: I guess you'd be right. LGTM 😄

docker/Dockerfile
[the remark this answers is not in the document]
jason-fox: Just a minute - if is this file `/entrypoint.sh`  how is it being copied to root?
The Git clone will copy it to `/docker/entrypoint.sh and I dont see a `RUN mv` in the Dockerfile

docker/Dockerfile
```diff
@@ -28,6 +33,7 @@ MAINTAINER FIWARE IoTAgent Team. Telefónica I+D
 # replace the whole RUN statement by the following COPY statement in your local source using :
 #
 # COPY . /opt/iotajson/
+# COPY ./docker/entrypoint.sh/
```
jason-fox: Missing space and extraneous slash - shouldn't it be:

```
# COPY . /docker/entrypoint.sh
```

Jagatjot: I do not understand Run mv, should ENTRYPOINT be  ["/docker/entrypoint.sh"] ??

jason-fox: Your `ENTRYPOINT` expects a file to be found in root `./entrypoint.sh`

The commented out copy statement copies the file to root (for a local build)

```Dockerfile
# COPY . /docker/entrypoint.sh
````

The line https://github.com/telefonicaid/iotagent-json/blob/0fbd002/docker/Dockerfile#L56

```bash
mv "${GITHUB_REPOSITORY}-${RELEASE}" /opt/iotajson && \
```

Moves all of the contents of the cloned repo into  `/opt/iotajson` - you therefore need to move `/opt/iotajson/docker/entrypoint.sh` to `/entrypoint.sh` otherwise the `Dockerfile` won't start.

The basic question is what currently happens when you build and run the `Dockerfile`?  How are you testing this change? How are you verifying that it works correctly?


For more background on the Dockerfile and how it **currently works** see:

* https://github.com/telefonicaid/iotagent-ul/pull/310
* https://github.com/telefonicaid/iotagent-json/issues/342

Jagatjot: When I am testing on my local the Dockerfile gives a following error-

Jagatjot: In the current Dockerfile, the ENTRYPOINT is ["bin/iotagent-json"] as all the contents of cloned repo are into /opt/iotajson therefore /bin/iotagent-json are also cloned into dis repo. Similarly when /docker/entrypoint.sh will be merged in master, it will also be part of cloned repo inside /opt/iotajson and the ENTRYPOINT ["/docker/entrypoint.sh"] will start. What do you think @jason-fox  ?

jason-fox: Whilst testing use build args - from the docs:

>* To download code from your own fork of the GitHub repository add the `GITHUB_ACCOUNT` and `GITHUB_REPOSITORY` arguments to the docker build command.
>
>```console
>docker build -t iot-agent . --build-arg GITHUB_ACCOUNT=<your account> --build-arg >GITHUB_REPOSITORY=<your repo>
>```

When deciding where the `entrypoint.sh` file needs to be follow common best practice.

Looking at 3 common databases (Mongo-DB, Postgres and MySQL) they all place the 
`docker-entrypoint.sh` relatively.

```Docker
ENTRYPOINT ["/docker/entrypoint.sh"]
```
is **always** looking for root

```Docker
ENTRYPOINT ["docker/entrypoint.sh"]
```

would be relative.

Jagatjot: Fixes https://github.com/telefonicaid/iotagent-json/pull/380/commits/9bc569e10bd0d96ad23c10d3b6bff090cb71f2df

jason-fox: I assume the comment in line 36 is no longer necessary and can be removed.

Jagatjot: Fixed by https://github.com/telefonicaid/iotagent-json/pull/380/commits/a320676d7372bc4db45c6baba501e010b1ccc6c7

jason-fox: :+1: LGTM

Jagatjot:
@jason-fox Thank you! @fgalan Could you please check as well ?

docker/entrypoint.sh
```diff
@@ -0,0 +1,47 @@
...
+    else
+        echo "***********************************************"
+        echo "INFO: IoT Agent Auth credentials have been overridden"
+        echo "***********************************************"
+    fi
+fi
+
+pm2 start /bin/iotagent-json
+
+/bin/bash
```
jason-fox: Delete this line - it is unnecessary

Jagatjot: Fixed by https://github.com/telefonicaid/iotagent-json/pull/380/commits/1483f7dd9e4b3835e6d3c4438dc03a8df4afceb4

Jagatjot: Which artifact is remaining?

jason-fox: The line `/bin/bash`

docker/entrypoint.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
```
jason-fox: The Shebang is incorrect for a Docker image based on Debian it should be
```bash
#!/bin/bash
```

docker/entrypoint.sh
Jagatjot: Fixed by https://github.com/telefonicaid/iotagent-json/pull/380/commits/183aabb981e1238c4a4b14d5a5f1b402d0a9c0ca

docker/Dockerfile
```diff
@@ -71,5 +76,5 @@ RUN \
 USER node
 ENV NODE_ENV=production
 
-ENTRYPOINT ["pm2-runtime", "bin/iotagent-json"]
+ENTRYPOINT ["docker/entrypoint.sh"]
```
jason-fox: This file has no execute permissions `chmod +x docker/entrypoint.sh`

docker/entrypoint.sh
```diff
@@ -0,0 +1,45 @@
...
+        echo "The connection to the Context broker fails"
+        echo "***********************************************"
+    else
+        echo "***********************************************"
+        echo "INFO: IoT Agent Auth credentials have been overridden"
+        echo "***********************************************"
+    fi
+fi
+
+pm2 start /bin/iotagent-json
```
jason-fox: This should start in no-daemon mode

docker/entrypoint.sh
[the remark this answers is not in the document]
jason-fox: The entrypoint file does not have execute permissions, amend the `RUN` statement above

```
	apt-get -y autoremove && \
	chmod +x docker/entrypoint.sh
```

docker/entrypoint.sh
Jagatjot: Fixed by f7afb0deb79e45e42a555ec47abcc9eda5f530bd

docker/entrypoint.sh
Jagatjot: Fixed by f7afb0deb79e45e42a555ec47abcc9eda5f530bd

docker/Dockerfile
Jagatjot: Fixed by 46dac69bbf5bdff53bd1c1b03cc287c713fbe38c

jason-fox:
@fgalan. This is basically there, but with all of the comments reviews and changes it's getting very hard to see if anything else is left. Since an equivalent change will be needed for all IoT Agents, I have ported the changes over to the equivalent [Ultralight](https://github.com/telefonicaid/iotagent-ul/pull/339) repo where a clean implementation can be reviewed further.

I would suggest you close this PR, continue your review over there and then backport the changes to this repo once completed.

Jagatjot:
@jason-fox This PR could also be reviewed easily by seeing the files changed section. But I guess you would have made the similar changes in Ultralight as done in this PR. Moreover I have been working on this issue from the past 10-12 days. So don't you think we need to merge the changes before closing this PR. What do you think @fgalan ?

fgalan:
I'll try to conduct a review today. Sorry for being late (I was waiting for the PR to "stabilize" to avoid to review a moving picture :).... I haven't review the 145 comment long discussion, so forgive in advance if I point out something already comment on provide feedback contrary to the one previously provided by others.

I understand this PR is in sync with PR https://github.com/telefonicaid/iotagent-ul/pull/339, isn't it? Thus probably is a good idea to review it together and (eventually) merge them in sync.

config.js
```diff
@@ -134,6 +134,15 @@ config.iota = {
      */
```
fgalan: As far as I understand, this PR is about modification in the docker part. The config.js file is not used by the new entry point file. Thus, let's keep it simple are avoid chaning the config.js file.

docker/README.md
```diff
@@ -150,3 +150,12 @@ COPY . /opt/iotajson/
 ```
 
 Full instructions can be found within the `Dockerfile` itself.
+
+### Docker Secrets
+
+As an alternative to passing sensitive information via environment variables, `_FILE` may be appended to some sensitive environment variables, causing the initialization script to load the values for those variables from files present in the container. In particular, this can be used to load passwords from Docker secrets stored in `/run/secrets/<secret_name>` files.
+
```
fgalan: I'd suggest to include an example here, in the same way Cygnus dockerhub README.md (https://hub.docker.com/r/telefonicaiot/fiware-cygnus/) does.

docker/entrypoint.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
fgalan: Missing license header, pls.

docker/entrypoint.sh
```diff
@@ -0,0 +1,45 @@
...
+        echo "The connection to the Context broker fails"
+        echo "***********************************************"
+    else
+        echo "***********************************************"
+        echo "INFO: IoT Agent Auth credentials have been overridden"
+        echo "***********************************************"
+    fi
+fi
+
+pm2-runtime start /bin/iotagent-json
```
fgalan: I understand this comes from:

```
ENTRYPOINT ["pm2-runtime", "bin/iotagent-json"]
```

Should it be (as shown now)

```
pm2-runtime start /bin/iotagent-json
```

or actually:

```
pm2-runtime /bin/iotagent-json
```
?

docker/entrypoint.sh
```diff
@@ -0,0 +1,45 @@
...
+
+if [[  -z "$IOTA_AUTH_ENABLED" ]]; then
+ echo "***********************************************"
+ echo "WARNING: It is recommended to enable authentication for secure connection"
+ echo "***********************************************"
+else
+    if  [[ -z "$IOTA_AUTH_USER" ]] || [ -z "$IOTA_AUTH_PASSWORD" ]]; then
+        echo "***********************************************"
+        echo "WARNING: Default IoT Agent Auth credentials have not been overridden"
+        echo "The connection to the Context broker fails"
```
fgalan: "The connection to the Context broker fails" -> "The connection to the Context broker will fail" (unsure, but I think the later is more correct)

docker/Dockerfile
```diff
@@ -52,8 +57,8 @@ RUN if [ "${DOWNLOAD}" = "latest" ] ; \
 	apt-get clean && \
 	apt-get remove -y unzip && \
 	apt-get -y autoremove
-
-
+
```
fgalan: Avoid adding extra whitespace with no purpose :)

docker/Dockerfile
```diff
@@ -66,10 +71,11 @@ RUN \
 	# Remove Git and clean apt cache
 	apt-get clean && \
 	apt-get remove -y git && \
-	apt-get -y autoremove
+	apt-get -y autoremove && \ 
+	chmod +x docker/entrypoint.sh
```
fgalan: Looking to Cygnus Dockerfile (https://github.com/telefonicaid/fiware-cygnus/blob/master/docker/cygnus-ngsi/Dockerfile) we don't need to chmod +x the entrypoint file there. Why this is needed here?

docker/Dockerfile
```diff
@@ -66,10 +71,11 @@ RUN \
...
 
 USER node
 ENV NODE_ENV=production
 
-ENTRYPOINT ["pm2-runtime", "bin/iotagent-json"]
+ENTRYPOINT ["docker/entrypoint.sh"]
```
fgalan: Looking to Cygnus Dockerfile (https://github.com/telefonicaid/fiware-cygnus/blob/master/docker/cygnus-ngsi/Dockerfile) we have: 

```
ENTRYPOINT ["/cygnus-entrypoint.sh"]
```

So why we need `docker/` here? It will be better to be homogeneous and have:

```
ENTRYPOINT ["/iotagent-json-entrypoint.sh"]
```

Jagatjot: @fgalan As explained in comment https://github.com/telefonicaid/iotagent-json/pull/380#discussion_r254773976 docker/entrypoint.sh was used.

docker/Dockerfile
Jagatjot: @fgalan As suggested in comment https://github.com/telefonicaid/iotagent-json/pull/380#discussion_r255063951 permission was added. What do you think ?

Jagatjot: When we run entrypoint.sh script in our local its asks for permission so we decided to add it in the Dockerfile itself.

fgalan: I understand that is you set the permission in the file itself in the repository you don't need to use chmod +x in the Dockerfile.

Note `ls` run in the repository working copy shows +x for that file

```
$ ls -l
total 88
-rw-r--r-- 1 fermin fermin 11366 feb 20 12:37 agent.conf
-rw-r--r-- 1 fermin fermin   159 sep 13  2017 cartodb_keys.conf
-rwxr-xr-x 1 fermin fermin 50304 dic 13 10:46 cygnus-entrypoint.sh
-rw-r--r-- 1 fermin fermin 12489 feb 20 12:37 Dockerfile
-rw-r--r-- 1 fermin fermin  1016 oct  9 18:51 README.md
```

This seems the way in which it is used in Cygnus, so please do the same.

jason-fox: @fgalan :

> we don't need to chmod +x the entrypoint file there. Why this is needed here?

The IoT Agents set a `USER` , cygnus less secure as it will be running using the default root user.

```Dockerfile
USER node
````

Maybe someone should raise this as a security issue on Cygnus?

The advantage of leaving in the `chmod` statement is that it is explicit. This way all files can be left as `-rw` by default and you're not relying on some access permissions flags. These permissions flags are hidden from the developer by default.

File: docker/Dockerfile (hunk 47 above)
Comment (jason-fox) [reply docker/Dockerfile]: Cygnus adds an unnecessary `COPY` statement to ensure the `cygnus-entrypoint.sh` file can be found in the root of the container. There are two copies of the same file within the image.

```Dockerfile
# Copy some files needed for starting cygnus-common
COPY cygnus-entrypoint.sh /
COPY agent.conf ${FLUME_HOME}/conf/

# Define the entry point
ENTRYPOINT ["/cygnus-entrypoint.sh"]
```

OK in the case of the Cygnus repository, this may make sense given there are multiple `Dockerfiles`, but for a simple repo with one `Dockerfile` such as this the duplication is unnecssary

config.js
jason-fox: @fgalan - I thought the `ENV` variables only overwrote *existing* attributes in the `config.js` file. If adding this is a necessary part of this PR since without it the Username/Password, secure flag & etc. won't get set.

Could you clarify please?

docker/Dockerfile
Jagatjot: @fgalan So the final conclusion is to add chmod +x I guess ?

docker/Dockerfile
Jagatjot: As entrypoint.sh script is written inside a docker folder so I feel ENTRYPOINT ["docker/entrypoint.sh"] would be a better option. What do you think @fgalan @jason-fox ?

config.js
fgalan: Not 100% sure but looking to coce like this https://github.com/telefonicaid/iotagent-node-lib/blob/master/lib/commonConfig.js#L144 it seems the env var processing logic is able to create what can be missing from config.js.

Another configuration would be great. @dcalvoalonso what do you think?

docker/Dockerfile
fgalan: Ok. No so important. NTC

However, I think is a common practice (maybe a bad habit? :) to run processes *inside* docker containers as root. However, if you feel this could be a security risk, maybe a PR fixing it (i.e. adding `USER` to the several Cygnus Dockerfiles plus usage of chmod +x on the entrypoint file) would be a good contribution.

docker/Dockerfile
fgalan: Ok. Makes sense. Let's keep entrypoint.sh file within docker/

NTC

fgalan:
@Jagatjot after closing today a couple of comments threads with NTC let me summarize the list of pending comments:

1. https://github.com/telefonicaid/iotagent-json/pull/380#discussion_r258419452 (this stills needs some feedback.... I hope @dcalvoalonso could provide some additional insigths)
2. https://github.com/telefonicaid/iotagent-json/pull/380#discussion_r258420219
3. https://github.com/telefonicaid/iotagent-json/pull/380#discussion_r258420304
4. https://github.com/telefonicaid/iotagent-json/pull/380#discussion_r258420973
5. https://github.com/telefonicaid/iotagent-json/pull/380#discussion_r258421229
6. https://github.com/telefonicaid/iotagent-json/pull/380#discussion_r258421416

In addition, you should upgrade this PR branch with a merge from master in order to solve the conflict currently shown in  `docker/Dockerfile`.

docker/entrypoint.sh
jason-fox: Personally I'd just remove that **second** line, because there is at least one circumstance where it wouldn't fail. Imagine this scenario (I know these are just edge cases, but...):

1) Override `IOTA_AUTH_ENABLED = true`  ✅ 
2) Replace the default `config.js` with an attached `VOLUME` holding the *real* username and /or password in plain text. 😮 

It's still not really **safe** IMHO, but it would work and therefore the connection wouldn't fail so the second line of the statement would be incorrect.

Maybe someone is holding their username/passwords as _configuration as code_ passing an encrypted file onto a machine and then mounting in within a docker instance?

docker/Dockerfile
jason-fox: The **principle of least privilege** applies here - *especially* regarding passwords.

* https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b

docker/entrypoint.sh
fgalan: What about

"The connection to the Context broker most probably fail"

?

Removing the line is also ok from my point of view.

docker/README.md
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/399c3fbda91e1b11cec196698faf4c457ffe8705

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/3a5d1c5f5d74cc9b34444041157ce0f03f3ef7f0

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/3a5d1c5f5d74cc9b34444041157ce0f03f3ef7f0

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/3a5d1c5f5d74cc9b34444041157ce0f03f3ef7f0

docker/Dockerfile
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/65a2722aa4334631cebb1b886366f44f180aced7

config.js
Jagatjot: Awaiting for response

fgalan: @jason-fox did a test and it works without that addition. However, as we has been dicussing in the "twin" PR maybe is a good idea to include the block but in a commented way. See details at https://github.com/telefonicaid/iotagent-ul/pull/339#discussion_r260704766

Thus, in conclusion, please add here the same block it as added it in IOTA-UL (copy-paste basically ;)

docker/README.md
fgalan: I have looked to and it seems line "Currently, this the `_FILE` suffix is supported for:" is duplicated. Please remove it at line 159.

docker/entrypoint.sh
fgalan: (Actually it seems it was 04ecf9c897835733a9715527617f84f1a790343c)

docker/entrypoint.sh
fgalan: I see now it is:

```
pm2-runtime opt/iotajson/bin/iotagent-json
```

I think is safer to use absolute pathname here. Could you change it to 

```
pm2-runtime /opt/iotajson/bin/iotagent-json
```

please?

fgalan: Actually, I have realized we have this in dockerfile:

```
WORKDIR /opt/iotajson
```

so it could be even better this way (closer to the original one before this PR):

```
pm2-runtime bin/iotagent-json
```

docker/entrypoint.sh
fgalan: (Actually it was fixed in commit 04ecf9c897835733a9715527617f84f1a790343c)

config.js
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/9a74a9a0480f11abfd56eae8461da68ebbb54585

docker/Dockerfile
jason-fox: LGTM

docker/README.md
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/ab8ded7ad6ea497c37bd3632425b675d3a8a550b

docker/entrypoint.sh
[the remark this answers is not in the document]
jason-fox: LGTM

docker/entrypoint.sh
Jagatjot: Fixed with https://github.com/telefonicaid/iotagent-json/pull/380/commits/20e74f925fbc7728990f3b4f2008e5a2247129de

dcalvoalonso:
LGTM! Very good job @Jagatjot and @jason-fox ! :)

fgalan:
Base branch changed to feature/366_pre_merge before merging.

PR #382, opened 2019-02-04
Update Node version

jason-fox:
Add security patches


docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG NODE_VERSION=8.12.0-slim
+ARG NODE_VERSION=8.15.0-slim
```
fgalan: Please tell about the change in  CHANGES_NEXT_RELEASE file. Something like this:

```
- Upgrade NodeJS version from 8.12.0 to 8.15.0 in Dockerfile due to security issues
```

jason-fox: Fixed #13cddf9

PR #392, opened 2019-03-31
Implements NGSIv2 context forwarding

dcalvoalonso:
Following the proposal of https://github.com/telefonicaid/iotagent-node-lib/pull/762#issuecomment-476645860, this PR adds unit tests for commands functionality using NGSIv2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "body-parser": "1.18.3",
     "dateformat": "3.0.3",
     "express": "~4.16.4",
-    "iotagent-node-lib": "git://github.com/telefonicaid/iotagent-node-lib.git#master",
+    "iotagent-node-lib": "git://github.com/dcalvoalonso/iotagent-node-lib.git#task/lazyAttsNgsiv2",
     "logops": "2.1.0",
     "mqtt": "2.18.8",
     "request": "2.88.0",
--- a/test/unit/ngsiv2/contextRequests/updateStatus2.json
+++ b/test/unit/ngsiv2/contextRequests/updateStatus2.json
@@ -1,5 +1,5 @@
 {
-	"actionType": "APPEND",
+	"actionType": "append",
 	"entities": [{
 		"id": "Second MQTT Device",
 		"type": "AnMQTTDevice",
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "body-parser": "1.18.3",
     "dateformat": "3.0.3",
     "express": "~4.16.4",
-    "iotagent-node-lib": "git://github.com/telefonicaid/iotagent-node-lib.git#master",
+    "iotagent-node-lib": "git://github.com/dcalvoalonso/iotagent-node-lib.git#task/lazyAttsNgsiv2",
```
dcalvoalonso: This change must be reverted once https://github.com/telefonicaid/iotagent-node-lib/pull/762 is merged and before merging the current PR.

test/unit/ngsiv2/contextRequests/updateStatus2.json
```diff
@@ -0,0 +1,15 @@
+{
+	"actionType": "APPEND",
```
fgalan: Action types are lowercase in NGSIv2 (uppercase are also allowed, but are deprecated). Thus, I'd suggest to use "append" here.

fgalan: ...which probably needs a modification in the agent or library code.

In fact, I wonder why append is used here. I mean, if the entity already exist (and it should be, otherwise a command cannot be send on it) it should be "update" instead of "append". However, if we are unsure of how this works, let's keep "append" (which is always "safer").

dcalvoalonso: Fixed in library with 2f7eff9e1bdf07be79dbae46eca9f2e0e0851a1a

dcalvoalonso: Fixed here with 669a9c3053f0af12d6947225cf9a9f863f8fdec5

fgalan:
Included in PR https://github.com/telefonicaid/iotagent-json/pull/430. Closing.

PR #393, opened 2019-04-02
Add missing protected Sensitive Parameters.

jason-fox:
This PR adds `IOTA_AUTH_CLIENT_ID` and  `IOTA_AUTH_CLIENT_SECRET` to the protected list.  Currently the `Dockerfile` **does support** them, but the existing version of iot-node-lib is missing the equivalent processing step. 

Related: https://github.com/telefonicaid/iotagent-node-lib/pull/767

Obviously assumes that a release with an updated `package.json` is imminent.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -3,133 +3,128 @@
 [![FIWARE IoT Agents](https://nexus.lab.fiware.org/repository/raw/public/badges/chapters/iot-agents.svg)](https://www.fiware.org/developers/catalogue/)
 [![](https://nexus.lab.fiware.org/repository/raw/public/badges/stackoverflow/iot-agents.svg)](https://stackoverflow.com/questions/tagged/fiware+iot)
 
-An Internet of Things Agent for a JSON based protocol (with
-[AMQP](https://www.amqp.org/), [HTTP](https://www.w3.org/Protocols/) and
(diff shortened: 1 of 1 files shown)


docker/README.md
```diff
@@ -3,141 +3,137 @@
 [![FIWARE IoT Agents](https://nexus.lab.fiware.org/repository/raw/public/badges/chapters/iot-agents.svg)](https://www.fiware.org/developers/catalogue/)
 [![](https://nexus.lab.fiware.org/repository/raw/public/badges/stackoverflow/iot-agents.svg)](https://stackoverflow.com/questions/tagged/fiware+iot)
 
-An Internet of Things Agent for a JSON based protocol (with
```
fgalan: Many of the changes in this .md seems to be of prettier nature... which wonders me a bit given that as far as I understand we already did the prettifi-cation in a past PR.

I feel curiosity. Why we are seeing again this kind of changes?

jason-fox: The existing `npm run prettier:text` task only updates documentation markdown - the command misses `Docker\README.md`

The husky hook covers every markdown file - hence the `Docker\README.md` is now properly formatted.

fgalan: Thanks for the clarification. I wonder if prettie:text should cover also Docker\README.md, as husky does, but probably it doesn't worth the time.

NTC

PR #448, opened 2019-11-07
Task/use amqp durable option

AlvaroVega:
https://www.rabbitmq.com/mqtt.html#durability
Durable = true is QoS level 1

### DIFF
--- a/lib/bindings/AMQPBinding.js
+++ b/lib/bindings/AMQPBinding.js
@@ -88,8 +88,13 @@ function start(callback) {
 
     var durable;
 
-    if (config.getConfig() && config.getConfig().amqp && config.getConfig().amqp.durable) {
-        durable = config.getConfig().amqp.durable;
+    if (
+        config.getConfig() &&
+        config.getConfig().amqp &&
+        config.getConfig().amqp.options &&
+        config.getConfig().amqp.options.durable
+    ) {
+        durable = config.getConfig().amqp.options.durable;
     } else {
         durable = constants.AMQP_DEFAULT_DURABLE;
     }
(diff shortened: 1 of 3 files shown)


lib/constants.js
```diff
@@ -56,6 +56,7 @@ module.exports = {
 
     AMQP_DEFAULT_EXCHANGE: 'amq.topic',
     AMQP_DEFAULT_QUEUE: 'iotaqueue',
+    AMQP_DEFAULT_DURABLE: true,
```
fgalan: I understand that the config.amqp.durable (= AMQP_DEFAULT_DURABLE env var) was already included in config.js and documentation but up to now the code didn't use it. Is my understanding correct?

AlvaroVega: Yes, that is

lib/bindings/AMQPBinding.js
```diff
@@ -86,6 +86,14 @@ function start(callback) {
         queue = constants.AMQP_DEFAULT_QUEUE;
     }
 
+    var durable;
+
+    if (config.getConfig() && config.getConfig().amqp && config.getConfig().amqp.durable) {
+        durable = config.getConfig().amqp.durable;
+    } else {
+        durable = constants.AMQP_DEFAULT_DURABLE;
+    }
```
fgalan: Looking how durable is being processed here I guess it should be (inside config.amqp):

```
    durable: true,
```

However, in config.js I see (https://github.com/telefonicaid/iotagent-json/blob/master/config.js#L100):

```
    options: { durable: true }
```

Maybe config.js should be fixed?

AlvaroVega: fixed in https://github.com/telefonicaid/iotagent-json/pull/448/commits/26774c007854239ca7bdb953dace103c10c4c714

lib/constants.js
fgalan: NTC

PR #454, opened 2019-11-20
check response obj before use it (http commands)

AlvaroVega:
time=2019-11-20T11:12:07.482Z | lvl=FATAL | corr=n/a | trans=n/a | op=IoTAgentNGSI.Global | srv=n/a | subsrv=n/a | msg=An unexpected exception has been raised. Ignoring: TypeError: Cannot read property 'statusCode' of undefined | comp=IoTAgent


lib/bindings/HTTPBinding.js
```diff
@@ -122,8 +122,8 @@ function executeCommand(apiKey, device, serializedPayload, callback) {
     }
 
     request(options, function(error, response, body) {
-        if (error || (response.statusCode !== 200 && response.statusCode !== 201)) {
-            callback(new errors.HTTPCommandResponseError(response.statusCode, error));
+        if (error || (response && response.statusCode !== 200 && response.statusCode !== 201)) {
+            callback(new errors.HTTPCommandResponseError(response ? response.statusCode : 400, error));
```
jason-fox: I think the logic should be error || !response ||(response.statusCode !== 200 && response.statusCode !== 201) - for the current predicate, the *400* clause will never fire. because if `resource` is *undefined* the `else` clause will run.

AlvaroVega: fixed in https://github.com/telefonicaid/iotagent-json/pull/454/commits/773c6884f46ee039ab5166497c7acff932f50baa

lib/bindings/HTTPBinding.js
```diff
@@ -121,9 +121,9 @@ function executeCommand(apiKey, device, serializedPayload, callback) {
         options.timeout = config.getConfig().http.timeout;
```
fgalan: CHANGEX_NEXT_RELEASE entry is missing? I'd suggest to use something similar to the one you used in IOTA-UL:

```
- Add: check response obj before use it handling http commands
```

lib/bindings/HTTPBinding.js
```diff
@@ -121,9 +121,9 @@ function executeCommand(apiKey, device, serializedPayload, callback) {
...
     }
 
-    request(options, function(error, response, body) {
-        if (error || (response.statusCode !== 200 && response.statusCode !== 201)) {
-            callback(new errors.HTTPCommandResponseError(response.statusCode, error));
+    request(options, function(error, response, body) {
```
fgalan: Trailing whitespace should be removed.

lib/bindings/HTTPBinding.js
AlvaroVega: fixed in https://github.com/telefonicaid/iotagent-json/pull/454/commits/60a22256040ef9c740d9e04aa0a69d73153fc971

lib/bindings/HTTPBinding.js
AlvaroVega: fixed in https://github.com/telefonicaid/iotagent-json/pull/454/commits/e18ea915a788a0b211a0c890113cad3f1e2b7206

lib/bindings/HTTPBinding.js
```diff
@@ -251,7 +251,7 @@ function sendConfigurationToDevice(apiKey, deviceId, results, callback) {
         return function(error, response, body) {
             if (error) {
                 innerCallback(error);
-            } else if (response.statusCode !== 200) {
+            } else if (response && response.statusCode !== 200) {
```
jason-fox: Not sure about this - what if there is no `response`? Do you really want to call the `innerCallback()` in L257 in this case or return an `errors.DeviceEndpointError` object

AlvaroVega: In that case innerCallback() will be call, like before the proposed change.

jason-fox: NTC- If that is what you want then this is fine by me. Just checking

PR #472, opened 2020-03-31
Add Southbound HTTPS support

jason-fox:
Align HTTP processing with changes previously made in the Ultralight Agent.

- cert, key (retrieved from files).
- add missing options to config
- update documentation.

### DIFF
--- a/config.js
+++ b/config.js
@@ -41,6 +41,42 @@ config.mqtt = {
      * consists of a series of parameters separated by commas.
      */
     thinkingThingsPlugin: true,
+
+    /**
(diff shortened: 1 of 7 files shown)


config.js
```diff
@@ -41,6 +41,42 @@ config.mqtt = {
      * consists of a series of parameters separated by commas.
```
fgalan, line 41: CHANGES_NEXT_RELEASE entry about the support to HTTPS should be added.

jason-fox: Fixed 6fd6ea9

Issue #426, opened 2019-09-04
Log always writes the same correlator and transaction id


manucarrace:
Tween issue as https://github.com/telefonicaid/iotagent-ul/issues/326

DeepshikhaNEC:
@manucarrace In the current scenario the corr id and trans id is associated with the domain.js which is created once when the application is started and active for the complete running application.
That’s why it is giving the same value every time.

As per my understanding to change the corr id and trans id, we need to clean the domain and create it again or re-write the both of the id’s associated to the domain. 
Please let me know your opinion. What are the scenarios on which the corr id and trans id need to change?

fgalan:
> What are the scenarios on which the corr id and trans id need to change?

As a general rule (applying to all FIWARE components logging this way), corrid is defined as the ID for an e2e flow, while transid is defined as the ID for a given transaction within the e2e flow. In the case of IOTAs I'd say that all e2e flows correspond to a single transaction, so both can be treated the same way (how it works right now? corrid is equal to transid in the same log message? that is a valid solution in the case of IOTAs, from my point of view).

corrid should change in each new transaction (i.e. each new measure received at southbound interface and each new request of the provision API in the northbound interface) and keep the same in all the log messages related with that transaction.

I hope this information clarifies the issue.

fgalan:
Moreover, I tried to solve this in the past, with PR https://github.com/telefonicaid/iotagent-ul/pull/327. The PR is old and I wasn't able to complete the work, but maybe is useful to you to continue the work from that point.

DeepshikhaNEC:
> Moreover, I tried to solve this in the past, with PR [telefonicaid/iotagent-ul#327](https://github.com/telefonicaid/iotagent-ul/pull/327). The PR is old and I wasn't able to complete the work, but maybe is useful to you to continue the work from that point.

@fgalan Hi, As per my understandanding it is already fixed for all type of http request and i need to fix it only for mqtt and amqp request. Please confirm me is it correct?

fgalan:
> @fgalan Hi, As per my understandanding it is already fixed for all type of http request and i need to fix it only for mqtt and amqp request. Please confirm me is it correct?

Yes. I think you understanding is correct. Thanks!

DeepshikhaNEC:
@fgalan Hi, I have fixed this issue #426. For this PR has been raised in iotagent-json as well as iotagent-node-lib. 

Request you to please confirm the same and if the PR is ok then please merge it in master branch.

[...]

PR #478, opened 2020-05-20
Fixed for iotagent-json issue #426

DeepshikhaNEC:
Fixed issue #426 for "Log always writes the same correlator and transaction id".

### DIFF
--- a/lib/commonBindings.js
+++ b/lib/commonBindings.js
@@ -26,6 +26,7 @@
 'use strict';
 
 var iotAgentLib = require('iotagent-node-lib'),
+    regenerateTransid = iotAgentLib.regenerateTransid,
     commandHandler = require('./commandHandler'),
     transportSelector = require('./transportSelector'),
     async = require('async'),
@@ -292,6 +293,7 @@ function messageHandler(topic, message, protocol) {
  * @param {Object} message      AMQP message body (Object or Buffer, depending on the value).
  */
 function amqpMessageHandler(topic, message) {
+    regenerateTransid(topic);
     messageHandler(topic, message, 'AMQP');
 }
 
@@ -303,6 +305,7 @@ function amqpMessageHandler(topic, message) {
  * @param {Object} message      MQTT message body (Object or Buffer, depending on the value).
  */
 function mqttMessageHandler(topic, message) {
+    regenerateTransid(topic);
     config.getLogger().debug(context, 'message topic: %s', topic);
     messageHandler(topic, message, 'MQTT');
 }
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "body-parser": "1.18.3",
     "dateformat": "3.0.3",
     "express": "~4.16.4",
-    "iotagent-node-lib": "git://github.com/telefonicaid/iotagent-node-lib.git#master",
+    "iotagent-node-lib": "git://github.com/DeepshikhaNEC/iotagent-node-lib.git#nodelib-issue/426",
     "logops": "2.1.0",
     "mqtt": "3.0.0",
     "request": "2.88.0",
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "body-parser": "1.18.3",
     "dateformat": "3.0.3",
     "express": "~4.16.4",
-    "iotagent-node-lib": "git://github.com/DeepshikhaNEC/iotagent-node-lib.git#nodelib-issue/426",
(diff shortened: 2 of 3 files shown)


lib/commonBindings.js
```diff
@@ -303,6 +305,7 @@ function amqpMessageHandler(topic, message) {
  * @param {Object} message      MQTT message body (Object or Buffer, depending on the value).
  */
 function mqttMessageHandler(topic, message) {
+    regenerateTransid(topic);
```
fgalan, line 308: For the sake of homogeneity, I think it would be a good idea to apply the same solution also to the HTTP tranport. That way, we will be using the same solution for all the transports supported by this agent (HTTP, MQTT and AMQP).

Don't worry about my PR https://github.com/telefonicaid/iotagent-ul/pull/327. That PR was old and only a draft. Your solution looks smarter, so I can close my PR at the end when yours get completed.

Thanks!

DeepshikhaNEC: > For the sake of homogeneity, I think it would be a good idea to apply the same solution also to the HTTP tranport. That way, we will be using the same solution for all the transports supported by this agent (HTTP, MQTT and AMQP).
> 
> Don't worry about my PR [telefonicaid/iotagent-ul#327](https://github.com/telefonicaid/iotagent-ul/pull/327). That PR was old and only a draft. Your solution looks smarter, so I can close my PR at the end when yours get completed.
> 
> Thanks!

@fgalan  Hi, I have checked the code for HTTP as well and it is already working fine. The functionality for MQTT and AMQP has been added in a way that they are maintaining the homogeneity.
However the code flow of http is different from the MQTT and AMQP protocol, So if it is required to sync the HTTP code as well, then there would be a huge amount of code will be changed. I suggest to keep HTTP code same as implemented by you in PR telefonicaid/iotagent-ul#327.
Please let me know your opinion.

fgalan: So,

* The HTTP code is working well (with regards to correlatorid) now (as in master branch, I mean) in IOTA-JSON but *not* in IOTA-UL
* The HTTP code will work well (with regards to correlatorid) in IOTA-UL when https://github.com/telefonicaid/iotagent-ul/pull/327 gets merged

Is my understanding correct?

DeepshikhaNEC: > So,
> 
> * The HTTP code is working well (with regards to correlatorid) now (as in master branch, I mean) in IOTA-JSON but _not_ in IOTA-UL
> * The HTTP code will work well (with regards to correlatorid) in IOTA-UL when [telefonicaid/iotagent-ul#327](https://github.com/telefonicaid/iotagent-ul/pull/327) gets merged
> 
> Is my understanding correct?

@fgalan Hi, The HTTP code is working well (with regards to correlatorid) for both  IOTA-JSON and IOTA-UL even without merging you PR telefonicaid/iotagent-ul#327.

Please find the attached screenshots(first one is for IOTA-UL and second one is for IOTA-JSON) according to which the new transaction id/ correlator id is generating uniquely after every new http request for both iotagent-json and iotagent-ul.

I suggest we can keep the same code (as in master branch) in both for http request as it is already working fine. 

And for MQTT and AMQP I have been raised my PR regards the iotagent-json issue #426.

totagent-ul-afterhttpreq-ss
iotagent-json-afterhttpreq-ss

package.json
```diff
@@ -60,7 +60,7 @@
     "body-parser": "1.18.3",
     "dateformat": "3.0.3",
     "express": "~4.16.4",
-    "iotagent-node-lib": "git://github.com/telefonicaid/iotagent-node-lib.git#master",
+    "iotagent-node-lib": "git://github.com/DeepshikhaNEC/iotagent-node-lib.git#nodelib-issue/426",
```
fgalan, line 63: PR https://github.com/telefonicaid/iotagent-node-lib/pull/864 has been just merged into master, so please revert this change.

lib/commonBindings.js
fgalan: > I suggest we can keep the same code (as in master branch) in both for http request as it is already working fine.

Great! We would follow your suggestion. My old PR goes to "on hold" state (and in the future it will be closed).

Thanks for the analysis work!

NTC

package.json
DeepshikhaNEC: > PR [telefonicaid/iotagent-node-lib#864](https://github.com/telefonicaid/iotagent-node-lib/pull/864) has been just merged into master, so please revert this change.

@fgalan  Hi, I have been reverted the changes of package.json

PR #481, opened 2020-06-29
Leading slashes mqtt topics

jmezzera:
Same as https://github.com/telefonicaid/iotagent-ul/pull/429

### DIFF
--- a/config.js
+++ b/config.js
@@ -104,7 +104,7 @@ config.mqtt = {
     /**
      * Whether to use slashes at the beginning of topic when sending or not
      */
-    avoidLeadingSlashes: false
+    avoidLeadingSlash: false
 };
 
 /**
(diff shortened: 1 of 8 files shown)


config.js
```diff
@@ -99,7 +99,12 @@ config.mqtt = {
     /**
      * Time to keep connection open between client and MQTT broker (default is 0 seconds)
      */
-    keepalive: 0
+    keepalive: 0,
+
+    /**
+     * Whether to use slashes at the beginning of topic when sending or not
+     */
+    avoidLeadingSlashes: false
```
fgalan, line 107:
```suggestion
    avoidLeadingSlash: false
```

(as in the IOTA-UL PR)

Once applied, I understand this change would need fixes in some other places.

fgalan: Or maybe it's a typo just in this place... (other parts of the PR use the singular form)

jmezzera: My bad. Definitely a typo. Thanks btw

fgalan: Fixed in df397c4

Issue #486, opened 2020-09-02
Error processing 0.0 messages


mrutid:
At function parseMessage

https://github.com/telefonicaid/iotagent-json/blob/449e3a2b05bb9f0dd452bf0be18f10be48f7c613/lib/commonBindings.js#L46

there is a problem determining the MEASURES 3 condition

Take into account that:
> JSON.parse('0')
0

is falsy.

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-json/pull/487

PR #487, opened 2020-09-02
Update commonBindings.js: fix evaluation of parseMessage 

AlvaroVega:
fix issue https://github.com/telefonicaid/iotagent-json/issues/486


CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+Fix evaluation of parseMessage (#486)
```
```suggestion
Error processing zero measures (#486)
```

I think this (or alternate improved wording) is better, as this way is closer to the user (parseMessage is an internal function).

fgalan (approved):
LGTM

PR #495, opened 2020-09-16
Fix/build hook

IvanHdzC:
This PR fixes docker hub building failures and pushes the two new images into the repository.

### DIFF
--- a/docker/hooks/post_push
+++ b/docker/hooks/post_push
@@ -1,0 +1,27 @@
+#!/bin/bash
+#
+# Copyright 2019 Telefonica Investigación y Desarrollo, S.A.U
+#
+# This file is part of iotagent-json
+#
+# iotagent-json is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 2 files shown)


docker/hooks/post_push
```diff
@@ -0,0 +1,27 @@
...
+#
+# You should have received a copy of the GNU Affero General Public
+# License along with iotagent-json. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by the GNU Affero General Public License please contact
+# with iot_support at tid dot es
+
+docker image push $IMAGE_NAME-pm2 
+
+docker image push $IMAGE_NAME-distroless
```
fgalan, line 25: I understand that the default image (the one for node-slim) is automatically pushed so there is no need of including it in posst_push. Is my understanding correct?

IvanHdzC: That's correct @fgalan. The tagged version is pushed by default. So there is no need to push it again.

fgalan: NTC

fgalan (approved):
LGTM

Issue #413, opened 2019-07-10
In mqtt set default keepalive parameter to default mqtt software value 60 seconds


cesarjorgemartinez:
In mqtt set default keepalive parameter to default mqtt software value 60 seconds

See: https://www.npmjs.com/package/mqtt or https://github.com/mqttjs/MQTT.js#connect

AlvaroVega:
Now this is a parameter, configurable in config.js or env var, so you can set the value that you want. Default is 0.

fgalan:
So probably this issue should be closed. @cesarjorgemartinez what do you think?

fgalan:
Fixed by PR https://github.com/telefonicaid/iotagent-json/pull/525

PR #525, opened 2021-01-12
set 60 default keepalive mqtt value

AlvaroVega:
fix https://github.com/telefonicaid/iotagent-json/issues/413

60 is default value for keepalive in python and nodejs mqtt libraries:

https://www.npmjs.com/package/mqtt#mqttclientstreambuilder-options


fgalan (approved):
LGTM

PR #485, opened 2020-08-18
Docker hardening

jason-fox:
This PR fixes several issues raised by [docker/docker-bench-security](https://github.com/docker/docker-bench-security)

- Add missing `HEALTHCHECK` - currently just a ping to `/iota/about`, but could be expanded
- Add standard metadata build `LABEL` (see [opencontainers.org](https://github.com/opencontainers/opencontainers.org))
- Remove **deprecated** `MAINTAINER` - replace with `LABEL` 
- Create two-stage build process  using a **Distroless** build:

     -    Remove [disto](https://medium.com/@dwdraju/distroless-is-for-security-if-not-for-size-6eac789f695f) 	
     -    Run as anonymous user 
     -    Set `NODE_ENV` to development/production

PR #526, opened 2021-01-14
Add docker image for debugging

AlvaroVega:
Including node inspection

Extend works from https://github.com/telefonicaid/iotagent-json/pull/485 done by @jason-fox

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -231,7 +231,7 @@ HEALTHCHECK --interval=30s --timeout=3s --start-period=10s \
 # - IOTA_AUTH_CLIENT_ID, IOTA_AUTH_CLIENT_SECRET - when using OAuth2 Security
 #
 ########################################################################################
-FROM node:${NODE_VERSION}-slim AS slim
+FROM node:${NODE_VERSION}-slim AS debug
 ARG GITHUB_ACCOUNT
 ARG GITHUB_REPOSITORY
 ARG NODE_VERSION
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -218,3 +218,41 @@ CMD ["node", "/opt/iotagent-json/bin/iotagent-json", "-- ", "config.js"]
 HEALTHCHECK --interval=30s --timeout=3s --start-period=10s \
    CMD ["npm", "run", "healthcheck"]
 
+
+########################################################################################
+#
+# This build stage creates a node-slim image for debugging.
+#
+# IMPORTANT: For production environments use Docker Secrets to protect values of the 
+# sensitive ENV variables defined below, by adding _FILE to the name of the relevant 
+# variable.
+#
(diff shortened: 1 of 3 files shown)


fgalan:
Should the hook files in https://github.com/telefonicaid/iotagent-json/tree/master/docker/hooks be modified to take into account this new image (unsure)?

fgalan:
Btw, I understand that a "mirror" PR should be done in IOTA-UL repo.

jason-fox:
Can you explain the reason a separate `debug` image is required. Is there any reason you couldn't just override the existing environment variable e.g.:

```console
docker pull iotagent-json
docker run iotagent-json -e NODE_ENV=development
```

If you really, really need to do this, the **default** Node Slim image should be the final image built so that this still works:
 
```console
docker build -t telefonicaiot/iotagent-json .
```

Even better would be to use a build  argument and pass it into the distroless and pm2 builds as well so that all base images can be built with debug or production.


```Dockerfile

ARG NODE_ENV=production

... etc

FROM node:${NODE_VERSION}-slim AS slim
ARG GITHUB_ACCOUNT
ARG GITHUB_REPOSITORY
ARG NODE_ENV

... etc

USER node
ENV NODE_ENV=${NODE_ENV}
```

The `docker/hooks/build` could easily rebuild your variants using cached layers and the build time for a user running the file directly wouldn't increase **at all**.


```console
docker build -t telefonicaiot/iotagent-json:debug . -build-arg NODE_ENV=development
docker push telefonicaiot/iotagent-json:debug
```

AlvaroVega:
@jason-fox debug image is using `--inspect-brk` node arg, that's the reason that I've see to do it. 
Anyway how can I use `--inspection` with a production env like, latest, distroless o pm2 image?

jason-fox:
Just add parameters to override the default `CMD` after the name of the image:


```console
docker run --name debug  telefonicaiot/iotagent-json node --inspect-brk=0.0.0.0 /opt/iotagent-json/bin/iotagent-json --  config.js
```

```text
Unable to find image 'telefonicaiot/iotagent-json:latest' locally
latest: Pulling from telefonicaiot/iotagent-json
8aff230071c9: Already exists 
11e9dbd0ef22: Already exists 
a4382e503884: Already exists 
7e969fceb79f: Already exists 
a54d76effca2: Already exists 
c676cf6b127f: Pull complete 
Digest: sha256:60530d8ceb4788847321cba76525c46cc7e57e3acfb68c1558819b541fd8e15e
Status: Downloaded newer image for telefonicaiot/iotagent-json:latest
Debugger listening on ws://0.0.0.0:9229/0f4c11b2-ba62-4775-ab64-f5ba3391d0fa
For help, see: https://nodejs.org/en/docs/inspector
```

or use the `command` attribute in a docker-compose

AlvaroVega:
Ok @jason-fox works to me without a new image. By the way, pm2 equivalent image could be achieve using this same solution. Thanks!

PR #558, opened 2021-04-16
Complete NGSIV2 Unit tests

mapedraza:
Updating NGSI V2 Unit tests with cases included on NGSI V1 before deleting them

### DIFF
--- a/test/unit/ngsiv2/commandsAmqp-test.js
+++ b/test/unit/ngsiv2/commandsAmqp-test.js
@@ -136,6 +136,12 @@ describe('AMQP Transport binding: commands', function() {
                 done();
             });
         });
+        it('should reply with the appropriate command information', function(done) {
+            request(commandOptions, function(error, response, body) {
+                should.exist(body);
+                done();
+            });
+        });
         it('should update the status in the Context Broker', function(done) {
             request(commandOptions, function(error, response, body) {
                 contextBrokerMock.done();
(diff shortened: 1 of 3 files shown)


test/unit/ngsiv2/commandsAmqp-test.js
```diff
@@ -136,6 +136,12 @@ describe('AMQP Transport binding: commands', function() {
                 done();
             });
         });
+        it('should reply with the appropriate command information', function(done) {
+            request(commandOptions, function(error, response, body) {
+                should.exist(body);
+                done();
+            });
+        });
```
mapedraza, line 144: Included tests from NGSI V1 version is failling in CI process

fgalan: I'll have a look to it.

mapedraza:
MQTT topics have to be changed to /APIKEY/DEVICE-ID/attrs to /**json**/APIKEY/DEVICE-ID/attrs

test/unit/ngsiv2/commandsAmqp-test.js
fgalan: I think I have realized why this works in NGSIv1 but not in NGSIv2.

In NGSIv1 all responses (even the ones for updates, like this one) has payload. With `should.exist(body)` you check that the response has a payload.

But in NGSIv2 you don't have response payload in the case of updates. Actually, the 204 No Content expectations check is asserting that behavior.

I suggest to remove this it() and incorporate the checking of no payload in the existing test above which should be something like this:

```
        it('should return a 204 OK without errors', function(done) {
            request(commandOptions, function(error, response, body) {
                should.not.exist(error);
                should.not.exist(body);
                response.statusCode.should.equal(204);
                done();
            });
        });
```

fgalan: Fixed in ee8be3c and ee8be3c

fgalan (approved):
LGTM

Issue #582, opened 2021-07-20
getCommand is returning bad content-type


AlvaroVega:
curl -i -X GET 'http://localhost:7897/iot/json?k=3ballfrt3t4s0e6yrucfe91cx&i=disp1&getCmd=1' 
HTTP/1.1 200 OK
X-Powered-By: Express
**Content-Type: text/html; charset=utf-8**
Content-Length: 20
ETag: W/"14-Nb0rFgrHgIzgTAOGKlZeMQdva9E"
Date: Tue, 20 Jul 2021 07:50:33 GMT
Connection: keep-alive
Keep-Alive: timeout=5

{"reset":{"s":"34"}}

AlvaroVega:
time=2021-07-20T07:55:33.932Z | lvl=DEBUG | corr=9b2c5301-2d0d-4adf-85d2-48d86f02992d | trans=9b2c5301-2d0d-4adf-85d2-48d86f02992d | op=IOTAJSON.HTTP.Binding | from=n/a | srv=n/a | subsrv=n/a | msg=Parsing payload [undefined] | comp=IoTAgent
time=2021-07-20T07:55:33.932Z | lvl=DEBUG | corr=9b2c5301-2d0d-4adf-85d2-48d86f02992d | trans=9b2c5301-2d0d-4adf-85d2-48d86f02992d | op=IOTAJSON.HTTP.Binding | from=n/a | srv=n/a | subsrv=n/a | msg=Parsed data: [undefined] | comp=IoTAgent
time=2021-07-20T07:55:33.932Z | lvl=DEBUG | corr=9b2c5301-2d0d-4adf-85d2-48d86f02992d | trans=9b2c5301-2d0d-4adf-85d2-48d86f02992d | op=IOTAJSON.HTTP.Binding | from=n/a | srv=n/a | subsrv=n/a | msg=Processing multiple HTTP measures for device [disp1] with apiKey [3ballfrt3t4s0e6yrucfe91cx] | comp=IoTAgent
[...]

PR #583, opened 2021-07-20
fix content-type for get command

AlvaroVega:
fix for https://github.com/telefonicaid/iotagent-json/issues/582

This PR uses accept header to build response content type for get commands

### DIFF
--- a/lib/bindings/HTTPBinding.js
+++ b/lib/bindings/HTTPBinding.js
@@ -470,15 +470,26 @@ function returnCommands(req, res, next) {
     if (req.query && req.query.getCmd === '1') {
         iotAgentLib.commandQueue(req.device.service, req.device.subservice, req.deviceId, function (error, list) {
             if (error || !list || list.count === 0) {
-                res.status(200).send('');
+                if (req.accepts('json')) {
+                    res.status(200).send({});
+                } else {
+                    res.status(200).send('');
+                }
             } else {
-                res.status(200).send(JSON.stringify(list.commands.map(parseCommand).reduce(concatCommand, {})));
-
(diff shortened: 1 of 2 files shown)


lib/bindings/HTTPBinding.js
```diff
@@ -470,15 +470,26 @@ function returnCommands(req, res, next) {
     if (req.query && req.query.getCmd === '1') {
```
fgalan, line 470: CNR entry missing, mentioning the fixed issue.

Is there a twin PR with a similar fix also for IOTA-UL? Or in that IOTA this is ok?

AlvaroVega: Iota UL https://github.com/telefonicaid/iotagent-json/issues/582#issuecomment-883229378 is returning a command in a a non json format, so IMHO there is no sense to allow return result in JSON

fgalan: Ok, application/json is not the right one in that case. However:

```
HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: text/html; charset=utf-8
Content-Length: 16
ETag: W/"10-qjxBJ6diMRAMIjnHPy48XmcadZE"
Date: Tue, 20 Jul 2021 10:21:28 GMT
Connection: keep-alive

disp3@close|s=36
```

should it be text/plain instead of text/html ?

fgalan: Fixed in independent PR https://github.com/telefonicaid/iotagent-ul/pull/507

fgalan (approved):
LGTM

PR #606, opened 2021-12-21
Add graceful shutdown

jason-fox:
This PR handles sigint and adds a npm start script.

When a shutdown signal is sent, we need to do graceful shutdown out of Node.js and
express to avoid side effects. like finishing active requests before closing server, clean up resources, db connections etc.

see: https://stackfame.com/node-express-graceful-shutdown

There are no unit tests, but if you start the app from the command-line and `ctrl-c` you'll see the handler being called.

### DIFF
--- a/lib/iotagent-json.js
+++ b/lib/iotagent-json.js
@@ -185,5 +185,20 @@ function stop(callback) {
     );
 }
 
+function handleShutdown(signal) {
+    config.getLogger().fatal(context, 'Received %s, starting shutdown processs', signal);
+    stop((err) => {
+        if (err) {
+            config.getLogger().error(context, err);
+            return process.exit(1);
+        }
+        process.exit(0);
+    });
+}
+
+process.on('SIGINT', handleShutdown);
+process.on('SIGTERM', handleShutdown);
+process.on('SIGHUP', handleShutdown);
+
 exports.start = start;
 exports.stop = stop;
--- a/lib/iotagent-json.js
+++ b/lib/iotagent-json.js
@@ -190,7 +190,7 @@ function stop(callback) {
  *
  */
 function handleShutdown(signal) {
-    config.getLogger().fatal(context, 'Received %s, starting shutdown processs', signal);
+    config.getLogger().info(context, 'Received %s, starting shutdown processs', signal);
     stop((err) => {
         if (err) {
             config.getLogger().error(context, err);
--- a/lib/iotagent-json.js
+++ b/lib/iotagent-json.js
@@ -185,6 +185,10 @@ function stop(callback) {
     );
 }
 
+/**
+ * Shuts down the IoT Agent in a graceful manner
+ *
+ */
 function handleShutdown(signal) {
     config.getLogger().fatal(context, 'Received %s, starting shutdown processs', signal);
     stop((err) => {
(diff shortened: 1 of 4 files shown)


jason-fox:
Obviously if this PR is approved, it can be copied across other IoT Agents.

lib/iotagent-json.js
```diff
@@ -185,5 +185,20 @@ function stop(callback) {
     );
```
fgalan, line 185: CHANGES_NEXT_RELEASE entry should be included, about the modifications in this PR.

package.json
```diff
@@ -27,6 +27,7 @@
     "lint:text": "textlint 'README.md' 'docs/*.md' 'docs/**/*.md'",
     "prettier": "prettier --config .prettierrc.json --write '**/**/**/*.js' '**/**/*.js' '**/*.js' '*.js'",
     "prettier:text": "prettier 'README.md' 'docs/*.md' 'docs/**/*.md' --no-config --tab-width 4 --print-width 120 --write --prose-wrap always",
+    "start": "node ./bin/iotagent-json",
```
fgalan, line 30: New npm target should be described in the corresponding .md (not remembering which one right now... but we have a description of all the different npm targets in some place)

fgalan:
> Obviously if this PR is approved, it can be copied across other IoT Agents.

Agreed. From our point of view, the following ones should be covered:

- [x] IOTA JSON (this PR)
- [ ] IOTA UL
- [ ] IOTA LWM2M
- [ ] IOTA Sigfox
- [ ] IOTA Manager

package.json
jason-fox: Fixed 0069f53

lib/iotagent-json.js
jason-fox: Fixed 0069f53

lib/iotagent-json.js
```diff
@@ -185,5 +185,24 @@ function stop(callback) {
...
 }
 
+/**
+ * Shuts down the IoT Agent in a graceful manner
+ *
+ */
+function handleShutdown(signal) {
+    config.getLogger().fatal(context, 'Received %s, starting shutdown processs', signal);
```
fgalan, line 193: Not sure 'fatal' would be the right one in this case... The 'fatal' loglevel is though for very critical fails and a ordered shutdown is not of that kind. Maybe 'info' is better.

@AlvaroVega @mapedraza what do you think?

jason-fox: You could get this SIGINT when running on Kubernetes. So it could be an outside process causing the IoT Agent to fail

fgalan: Note Orion is using INFO level for the same purpose:

https://github.com/telefonicaid/fiware-orion/blob/78db2d16ff7052cf8e46c0db4c69dc106172fc3a/src/app/contextBroker/contextBroker.cpp#L530

AlvaroVega: Info is enough from my point of view

jason-fox: Fixed b86a366

fgalan (approved):
LGTM. Thanks!

Waiting to have a full check-in in https://github.com/telefonicaid/iotagent-json/pull/606#issuecomment-998746073 to merge the 5 PRs at the same time.

fgalan:
All related PRs has been merged. Thanks!

Issue #610, opened 2022-01-12
log to info this message


AlvaroVega:
time=2022-01-12T10:52:59.908Z | lvl=DEBUG | corr=de682547-5bad-4cf6-9b73-0ee05091ed8e | trans=de682547-5bad-4cf6-9b73-0ee05091ed8e | op=IOTAJSON.MQTT.Binding | from=n/a | srv=n/a | subsrv=n/a | msg=Successfully subscribed to the following topics:
["/json/h0ydq/+/attrs/+"]

PR #615, opened 2022-01-13
log as info subscribed topics

AlvaroVega:
issue https://github.com/telefonicaid/iotagent-json/issues/610

no CNR update is needed


fgalan (approved):
LGTM

fgalan:
What about IOTA-UL? Have a similar log sentence?

In that case maybe we should create now that PR and merge at the same time, for not forgetting it...

fgalan:
> What about IOTA-UL? Have a similar log sentence?
> 
> In that case maybe we should create now that PR and merge at the same time, for not forgetting it...

I have just seen PR https://github.com/telefonicaid/iotagent-ul/pull/531 :)

PR #633, opened 2022-03-16
[WIP] remove support to send mqtt to agent wihtout protocol

AlvaroVega:
Sooner or later we will remove support to send to agent using mqtt topics without protocol.
But keeping from agent to device.


fgalan:
A twin PR should be create in IOTA-UL repository?

PR #655, opened 2022-04-27
Add explicit protocol note for MQTT topics

jason-fox:
Equivalent to https://github.com/telefonicaid/iotagent-ul/pull/557

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,4 +4,6 @@ mkdocs:
   configuration: mkdocs.yml
 
 python:
-  version: 3.6
+  version: 3.8
+  install:
+    - requirements: docs/requirements.txt
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,0 +1,4 @@
+mkdocs==1.2.3
+Pygments==2.9.0
+Markdown==3.3.4
+jinja2==3.0.0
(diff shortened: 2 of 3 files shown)


.readthedocs.yml
```diff
@@ -4,4 +4,6 @@ mkdocs:
   configuration: mkdocs.yml
 
 python:
-  version: 3.6
+  version: 3.8
```
fgalan, line 7: Currently this project has the following in the settings part at RTD admin web:

imagen

I understand that the settings in .readthedocs.yml override the ones in the RTD admin web. Is this correct? Or it is a not related setting?

jason-fox: Yes, the project file will override the defaults.  

Is anyone still using Python 2?

fgalan: > Is anyone still using Python 2?

Nobody *should* use Python 2, but for sure a lot of places (legacy, mostly) are still using it.

Not sure about reliable usage statistics on Python versions, but if we look to:

* https://w3techs.com/technologies/details/pl-python/2 (0.5%)
* https://w3techs.com/technologies/details/pl-python/3 (0.8%)

Based on that, Python 3 would be 1.6 times more used that Python 2... which would mean that a lot of legacy is still there today (a good ratio would be 10 or even 100 times ;)

fgalan: NTC, anycase

fgalan (approved):
LGTM

PR #666, opened 2022-05-16
Remove useless code

Emmanuel-Nau:


fgalan:
Has https://github.com/telefonicaid/iotagent-ul/ the same problem? Maybe it has been also fixed in a PR?

fgalan:
Btw, unit tests seem to be failing. That needs to be investigated and fixed before this PR gets merged.

Emmanuel-Nau:
Yes the tests are failing, but I think it was the case before this PR. For sure, we cannot merge this while it's failing. 
In iotagent-ul, the code is different and it is usefull.
Maybe you want to check this first, to make the change on this agent in place of juste removing the code ?

Emmanuel-Nau:
Ok, I think I saw what the problem is. 
The tests behave like they should, and they are proven this code is not useless! 
In fact it is just the use of the dataWithTimeStampArr that seems useless.
Because timeStampData is a reference to a measure object in the payload array, it's modification (adding the timestamp attribute) is reflected in the payload. 
So, I will update my code to only remove the dataWithTimeStampArr if you do not planned to use it. 
Sorry for my misreading of the code.

AlvaroVega:
Now it seems equivalent to:
https://github.com/telefonicaid/iotagent-ul/blob/1fbcff24dd7dc68e25b914c9805caa214715b8fe/lib/bindings/HTTPBindings.js#L96-L104

AlvaroVega (approved):
LGTM but changelog should be updated

fgalan (approved):
LGTM

Issue #679, opened 2022-07-20
loop error if none mosquitto config is provided


AlvaroVega:
When iotagent is started without mosquitto (i.e. without any IOTA_MQTT_XXXX option)

Several errors are continuously logged:

time=2022-07-20T07:40:17.111Z | lvl=FATAL | corr=n/a | trans=n/a | op=n/a | from=n/a | srv=n/a | subsrv=n/a | msg=GLOBAL-002: Couldn't connect with MQTT broker: {"errno":-111,"code":"ECONNREFUSED","syscall":"connect","address":"127.0.0.1","port":1883} | comp=IoTAgent
time=2022-07-20T07:40:17.111Z | lvl=FATAL | corr=n/a | trans=n/a | op=IoTAgentNGSI.Global | from=n/a | srv=n/a | subsrv=n/a | msg=An unexpected exception has been raised. Ignoring: Error: Callback was already called.
    at /opt/iotagent-json/node_modules/async/dist/async.js:969:32
    at MqttClient.<anonymous> (/opt/iotagent-json/lib/bindings/MQTTBinding.js:285:17)
    at MqttClient.emit (node:events:539:35)
    at MqttClient.emit (node:domain:475:12)
    at Socket.streamErrorHandler (/opt/iotagent-json/node_modules/mqtt/lib/client.js:460:12)
    at Socket.emit (node:events:527:28)
    at Socket.emit (node:domain:475:12)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) | comp=IoTAgent

AlvaroVega:
This bevaviour was introduced from 1.22.0 https://github.com/telefonicaid/iotagent-json/releases/tag/1.22.0

fgalan:
Fixed in PR https://github.com/telefonicaid/iotagent-json/pull/682

PR #682, opened 2022-07-21
 handle errors mqttt binding ending client

AlvaroVega:
fix for https://github.com/telefonicaid/iotagent-json/issues/679

This way, with default config, 5 tries for connect will be done before stop when an error of conection (ENOTFOUND, ECONNREFUSED or similar) is catched

Twin PR for UL agent is expected


fgalan (approved):
LGTM

PR #689, opened 2022-09-14
Feature/amqps protocol

StWiemann:
Since we are going to use AMQP in a future project I changed the URI-generation a bit to enable TLS-connections. 
It does not yet support client-certificates, since that isn't strictly needed in my use-case but I thought maybe somebody else might be interested in AMQPS. 

I did it analogous to mqtts, even though that might confuse some people due to AMQP-Protocols 0.9.1 vs 1.0 being a very different thing. This still uses 0.9.1, nothing changed there.

The new Env would be: 
`- IOTA_AMQP_PROTOCOL=amqp` or `amqps`

Don't really know if the testing-part works as intended. Basically just c/p'ed from MQTT_PROTOCOL...

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   },
   "keywords": [],
   "dependencies": {
-    "amqplib": "~0.5.1",
+    "amqplib": "~0.10.3",
     "async": "2.6.4",
     "body-parser": "1.20.0",
     "dateformat": "3.0.3",
--- a/docs/installationguide.md
+++ b/docs/installationguide.md
@@ -136,8 +136,8 @@ These are the currently available MQTT configuration options:
     order versions) or without the slash. See
     [discussion](https://github.com/telefonicaid/iotagent-node-lib/issues/866).
 -   **clean**: this flag is by default true, set to false to receive QoS 1 and 2 messages while offline.
--   **clientId**: string ID which identifies client in mqtt broker. By default is using a string composed by a fixed prefix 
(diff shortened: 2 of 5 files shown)


mapedraza:
Thank you for your contribution @StWiemann 

An entry to [CHANGES_NEXT_RELEASE](https://github.com/telefonicaid/iotagent-json/blob/master/CHANGES_NEXT_RELEASE) file is required before merging. 

Could you please open a twin PR on https://github.com/telefonicaid/iotagent-ul repo?

package.json
```diff
@@ -58,7 +58,7 @@
   },
   "keywords": [],
   "dependencies": {
-    "amqplib": "~0.5.1",
+    "amqplib": "~0.10.3",
```
fgalan, line 61: This should be also mentioned in the CHANGES_NEXT_RELEASE file. Something like this:

```
- Upgrade amqplib dependency from ~0.5.1 to ~0.10.3
```

fgalan:
> Thank you for your contribution @StWiemann
> 
> An entry to [CHANGES_NEXT_RELEASE](https://github.com/telefonicaid/iotagent-json/blob/master/CHANGES_NEXT_RELEASE) file is required before merging.
> 
> Could you please open a twin PR on https://github.com/telefonicaid/iotagent-ul repo?

@StWiemann did you take into account this feedback provided by @mapedraza , please? Thanks!

PR #700, opened 2023-03-02
use node _lib refactor

AlvaroVega:
PR to test agent with new iota node lib:
If timestamp is provided as measure then no metadata with timestamp is added to attributes.
Compression about timestamp is always checked in iotagent-node-lib, so always is uncompressed

iota-node-lib deb should be set to #master before merge

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,0 +1,1 @@
+- Align tests with iota-node-lib (v3.0) refactor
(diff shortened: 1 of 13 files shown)


fgalan:
> iota-node-lib deb should be set to #master before merge

Once merce PR https://github.com/telefonicaid/iotagent-node-lib/pull/1314, changed back to master in commit [4e98c7e](https://github.com/telefonicaid/iotagent-json/pull/700/commits/4e98c7e006b94696be93972db27595c422484f40)

fgalan (approved):
LGTM

PR #712, opened 2023-04-25
align tests with metadata timestamp when timestamp is measure fix

AlvaroVega:
related https://github.com/telefonicaid/iotagent-node-lib/pull/1344


CHANGES_NEXT_RELEASE
```diff
@@ -1,2 +1,3 @@
+- Align test with fix about propagate TimeInstant to all metadata attributes when TimeInstant is provided as measure
```
```suggestion
- Fix: align test with fix about propagate TimeInstant to all metadata attributes when TimeInstant is provided as measure
```

fgalan (approved):
LGTM

PR #730, opened 2023-06-21
update patch -> post according default value for appendMode is true

AlvaroVega:
related https://github.com/telefonicaid/iotagent-node-lib/pull/1386


mapedraza (approved):
LGTM

PR #785, opened 2023-11-24
Fix skipping tests at should level

mapedraza:
This PR fix skipping tests at should level (why now is not passing tests at master)

Twin PR: https://github.com/telefonicaid/iotagent-node-lib/pull/1535


fgalan (approved):
LGTM

PR #793, opened 2023-12-13
use transport and endpoint from group

AlvaroVega:
Before this PR when a device is provisioned default transport where HTTP if no other one is set at device provisioning time. After this PR default transport could be transport defined at Group Level.

Moreover polling or push behavior has into account possible endpoint defined at Group level.

- [ ] Update doc
- [ ] Add tests if apply


AlvaroVega:
closed by https://github.com/telefonicaid/iotagent-json/pull/794

PR #794, opened 2023-12-15
usage transport endpoint from group

AlvaroVega:
Before this PR when a device is provisioned default transport where HTTP if no other one is set at device provisioning time. After this PR default transport could be transport defined at Group Level.

Moreover polling or push behavior has into account possible endpoint defined at Group level.

- [x] Update doc
- [x] Add tests (about command groups for HTTP and MQTT transports)
- [x] Remove local implementation of `findOrCreate` 

Needs https://github.com/telefonicaid/iotagent-node-lib/pull/1543 and 
https://github.com/telefonicaid/iotagent-node-lib/pull/1547

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,2 +1,2 @@
 - Fix: protect access to multimeasure array
-- ADD: check and usage endpoint and transport from Group level when commands
+- Add: check and usage endpoint and transport from Group level when commands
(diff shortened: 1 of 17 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -1 +1,2 @@
 - Fix: protect access to multimeasure array
+- ADD: check and usage endpoint and transport from Group level when commands
```
fgalan, line 2:
```suggestion
- Add: check and usage endpoint and transport from Group level when commands
```

fgalan:
Twin PR https://github.com/telefonicaid/iotagent-ul/pull/656

fgalan (approved):
LGTM

PR #802, opened 2024-01-26
Add jexl example to ngsi payload

danielvillalbamota:


docs/usermanual.md
```diff
@@ -264,6 +264,17 @@ Some additional considerations to take into account:
     are include as NGSI-v2 metadata in the entity corresponding to the measure at Context Broker. Note IOTA doesn't
     provide the `type` for that metadata, so the Context Broker applies
     [a default type based in the metadata `value` JSON type](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#partial-representations).
+-   JEXL expression take place after key-value payload transformation, i.e. (1) NGSI-v2 batch update example after key-value tranformation would be:
```
mapedraza, line 267: What do you think about reformulate it in this way:

Only attributes value are available in JEXL context. In other words, it is not possible to access to attribute type or metadata. In order to address to the attribute value when using it in a JEXL expression, it is only required to use the attribute name. For example, in order to use `attr2` value from pervious example, the expression should be something like this: `"expression":"'This is my'+attr2"`

Makes sense?

PR #809, opened 2024-01-30
Task/include entity id and type as measures when ngsi payload

AlvaroVega:


CHANGES_NEXT_RELEASE
```diff
@@ -1 +1,2 @@
 - Fix: include entity id and entity type as measure value when measure is an ngsi (both NGSIv2 and NGSI-LD) payload
```
```suggestion
- Fix: include entity id and entity type as measure_id and measure_type at CB when measure is an ngsi (both NGSIv2 and NGSI-LD) payload
```

fgalan (approved):
LGTM

Issue #808, opened 2024-01-30
Support \t in Content-Type header


rg2011:
### IoT Agent JSON version the issue has been seen with

3.1.0

### Bound or port used (API interaction)

Southbound (Devices data API)

### NGSI version

NGSIv2

### Are you running a container?

Yes, I am using a contaner (Docker, Kubernetes...)

### Image type

normal

### Expected behaviour you didn't see

This is a follow-up to issue https://github.com/telefonicaid/iotagent-json/issues/759. The XML service integration still fails because the client includes a `"\t"` character in the Content-Type header. 

When sending a measure with content-type `"application/soap+xml;\tcharset=utf-8"` to `/iot/json/attrs/<name of attrib>`, the request should be accepted according to https://github.com/telefonicaid/iotagent-json/pull/771

### Unexpected behaviour you saw

The IoT Agent replies inmediately with "400 Bad Content".

This is because the [`type-is` module](https://www.npmjs.com/package/type-is), used [internally by express](https://github.com/expressjs/express/blob/2a00da2067b7017f769c9100205a2a5f267a884b/lib/request.js#L278-L290), fails to match this content-type to `application/soap+xml`:

```javascript
$ node
Welcome to Node.js v20.10.0.
Type ".help" for more information.
> const typeis = require('type-is').is
undefined
> const ctype = "application/soap+xml; action=\"urn:notificarEvento\"; \tcharset=utf-8"
undefined
> typeis(ctype, ["application/soap+xml"])
false
> typeis(ctype, ctype)
false
> 
```

This causes the following checks to fail:

https://github.com/telefonicaid/iotagent-json/blob/db402996729536d42e94ce70101bdfd896a793ea/lib/bindings/HTTPBinding.js#L75-L83

https://github.com/telefonicaid/iotagent-json/blob/db402996729536d42e94ce70101bdfd896a793ea/lib/bindings/HTTPBinding.js#L109-L115

And also an [internal test in the `bodyparser-xml` dependency](https://github.com/fiznool/body-parser-xml/blob/fed5c1849979d9c8695f24971c2a6f0f8dd3b885/index.js#L5), which performs its own content-type check, matching to `['*/xml', '+xml']` by default:

Removing the `\t` makes `type-is` recognize the content-type properly:

```javascript
> const ctype2 = "application/soap+xml; action=\"urn:notificarEvento\"; charset=utf-8"
undefined
> typeis(ctype2, ["application/soap+xml"])
'application/soap+xml'
[...]

PR #814, opened 2024-02-05
Add support for `\t` in content-type header

rg2011:
This PR is a workaround for issue #808 . The main aspects of the workaround are:

- Replace `req.is` with `typeis.is`. [type-is](https://github.com/jshttp/type-is) is a [dependency of `express`](https://github.com/expressjs/express/blob/2a00da2067b7017f769c9100205a2a5f267a884b/package.json#L59), so this does not add new dependencies.
- Change the call `bodyParser.xml({ ... })` to add the option `bodyParser.xml({ type: () => true, ... }`, which overrides [the internal check of the content-type header that `body-parser-xml` does](https://github.com/fiznool/body-parser-xml/blob/fed5c1849979d9c8695f24971c2a6f0f8dd3b885/index.js#L5C7-L17).
- Check the content-type header before parsing the body, by moving the content-type checking code from the `parseBody` function to a new middleware `checkPostContentType` that is called before parsing the body.


rg2011:
replaced by #817

PR #817, opened 2024-02-22
add argo binding

rg2011:
Adds support for internal SOAP service

### DIFF
--- a/lib/bindings/ARGOBinding.js
+++ b/lib/bindings/ARGOBinding.js
@@ -164,7 +164,7 @@ function findProperty(item, property, depth) {
             // In that case, the actual value should be in
             // the property named "_" of the object.
             if (typeof value === 'object' && !(value instanceof String)) {
-                if (!value || !Object.hasOwn(value, '_')) {
+                if (!value || !_.has(value, '_')) {
                     return ''; // empty value found
                 }
                 value = value._;
--- a/lib/bindings/ARGOBinding.js
+++ b/lib/bindings/ARGOBinding.js
@@ -1,0 +1,548 @@
+/*
+ * Copyright 2016 Telefonica Investigación y Desarrollo, S.A.U
+ *
+ * This file is part of iotagent-json
+ *
+ * iotagent-json is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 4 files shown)


rg2011:
> I'd suggest to add this entry to CHANGES_NEXT_RELEASE file:
> 
> ```
> - Add: ARGO binding
> ```

fixed

lib/bindings/ARGOBinding.js
```diff
@@ -0,0 +1,548 @@
...
+/**
+ * Middleware to add context to request
+ */
+function addLogContext(req, res, next) {
+    const trans = uuid.v4();
+    const corr = req.get('fiware-correlator') || trans;
+    req.logContext = {
+        path: req.path,
+        op: req.url,
+        //start: Date.now(),
```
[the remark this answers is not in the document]
rg2011: renoved

lib/bindings/ARGOBinding.js
```diff
@@ -0,0 +1,547 @@
...
+    </soap:Envelope>`);
+}
+
+/**
+ * Start the ARGO binding
+ */
+function start(callback) {
+    const logger = configService.getLogger();
+    const upstreamConf = configService.getConfig();
+    const localConf = processEnvironmentVariables(logger, {
```
AlvaroVega, line 422: Maybe some configuration (host, port, remote_url and a disable flag) could be added to https://github.com/telefonicaid/iotagent-json/blob/master/lib/configService.js

fgalan (approved):
LGTM with regards to my comments

@AlvaroVega has also done some comment, so the PR needs first a LGTM on his side before be merged.

lib/bindings/ARGOBinding.js
rg2011: the host, port and disable flag do exist, but I kept them all inside this file to make the service self-contained and "pluggable". In the future you could deprecate this feature from the iota-json anytime (e.g. because it migrates to a dedicated iota), just by removing this file and its tests.

But if you feel it is better to move some of the constants, flags or errors to the shared files, I am ok with that too.

AlvaroVega (approved):
LGTM

lib/bindings/ARGOBinding.js
fgalan: NTC at the end.

PR #822, opened 2024-04-11
Task/add test cases about multimeasure with timeinstant mapped attribute

AlvaroVega:
- [x] Non mapped TimeInstant, just a measure case
- [x] Mapped attribute to TimeInstant case: needs https://github.com/telefonicaid/iotagent-node-lib/pull/1593
- [x] Mapped Timestamp attribute to TimeInstant case


fgalan (approved):
LGTM

PR #847, opened 2024-11-28
add test about ngsiv2 measure with TimeInstant

AlvaroVega:
~~related and needs https://github.com/telefonicaid/iotagent-node-lib/pull/1672~~

This example only relies on device/group payloadType, which should be defined as "ngsiv2" for this case as https://github.com/telefonicaid/iotagent-json/blob/b00d69d64b1bb4c431b4e1724b69081b3256803d/test/unit/ngsiv2/deviceProvisioning/provisionDeviceHTTP2.json#L10


fgalan (approved):
LGTM

PR #859, opened 2025-01-29
update test about lastMeasure and add test about oldCtxt in jexl usage

AlvaroVega:
needs https://github.com/telefonicaid/iotagent-node-lib/pull/1691


fgalan (approved):
LGTM

PR #870, opened 2025-03-24
Update usermanual.md about expressions in commands:

AlvaroVega:
- [x] command issues when is defined at group level and then modified.
- [x] Add elements in ctxt: for comand and endppoint expression
- [x] Add examples of usage

### DIFF
--- a/docs/usermanual.md
+++ b/docs/usermanual.md
@@ -524,6 +524,8 @@ Content-type: application/json
 }
 ```
 
+-  Commands can be defined at group level as well as device level. When a command is just defined at group level command is copied to device and futher modifications are not reflected to device level. So if a command was defined at group level and then modified just at group level, devide command should be modified manually in order to get the same modification.
+
 #### Time processing
 
 HTTP bindig is returning in a HTTP header named `X-Processing-Time` processing time (in milliseconds) expended by current HTTP measure
--- a/docs/usermanual.md
+++ b/docs/usermanual.md
@@ -521,7 +521,7 @@ Content-type: application/json
   "move": 20
 }
 ```
--   Commands can be defined at the group level as well as the device level. When a command is defined at the group level, the command is copied to the device, and further modifications are not reflected at the device level. So if a command was defined at the group level and then modified only at the group level, the device command should be modified manually to reflect the same changes.
+-   Commands can be defined at the group level as well as the device level. When a command is defined at the group level, the command is copied to the device, and further modifications are not reflected at the device level. So if a command was defined at the group level and then modified only at the group level, the device command should be modified manually to reflect the same changes. In other words, the command at group level could be seen as a *template* that it is *instantiated* each time a new device on that group is created.
 
 #### Time processing
 
--- a/docs/usermanual.md
+++ b/docs/usermanual.md
@@ -856,10 +856,16 @@ expression of the **command** as well as the **device endpoint** when apply. In 
 -   `attribute`: attribute name and value involved in the command.
 
 This way an example for use jexl expressions for **endpoint** could be:
-
 ```json
 {
     "endpoint": "'https://myendpoint' + '/' + service + '/' + subservice + '/' + id + '/' + type"
+}
+```
+
+And following with the same command example usage, an example for use jexl expression with commands could be:
+```json
+{
+    "set": id + '_reset_' + reset
 }
 ```
 
--- a/docs/usermanual.md
+++ b/docs/usermanual.md
@@ -452,6 +452,7 @@ E.g.:
 #### Commands
 
 All the interations between IotAgent and ContextBroker related to comamnds are described in:
+
 -   [Theory: Scenario 3: commands](https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/devel/northboundinteractions.md#scenario-3-commands)
 -   [Practice: Scenario 3: commands - happy path](https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/devel/northboundinteractions.md#scenario-3-commands-happy-path)
 -   [Practice: Scenario 3: commands - error](https://github.com/telefonicaid/iotagent-node-lib/blob/master/doc/devel/northboundinteractions.md#scenario-3-commands-error)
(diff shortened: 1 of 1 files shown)


docs/usermanual.md
```diff
@@ -523,6 +521,7 @@ Content-type: application/json
   "move": 20
 }
 ```
+-   Commands can be defined at the group level as well as the device level. When a command is defined at the group level, the command is copied to the device, and further modifications are not reflected at the device level. So if a command was defined at the group level and then modified only at the group level, the device command should be modified manually to reflect the same changes.
```
fgalan, line 524:
```suggestion
-   Commands can be defined at the group level as well as the device level. When a command is defined at the group level, the command is copied to the device, and further modifications are not reflected at the device level. So if a command was defined at the group level and then modified only at the group level, the device command should be modified manually to reflect the same changes. In other words, the command at group level could be seen as a *template* that it is *instantiated* each time a new device on that group is created.
```

In case this clarification/addition helps. Otherwise, ignore it with NTC.

docs/usermanual.md
```diff
@@ -843,6 +842,30 @@ the example case the command will be:
 }
 ```
 
+The agent provides additional information, hereafter referred as **context**, in order to be used to evaluate the
```
fgalan, line 845:
```suggestion
The agent provides additional information, hereafter referred as **JEXL context**, in order to be used to evaluate the
```

To be clear it is not the sames the context information managed by ContextBroker, sometimes referred as **NGSI context**.

docs/usermanual.md
```diff
@@ -843,6 +842,30 @@ the example case the command will be:
...
+expression of the **command** as well as the **device endpoint** when apply. In all cases the following data is available to all expressions:
```
fgalan, line 846:
```suggestion
expression of the **command** as well as the **device endpoint** when apply. In all cases the following data is available to all expressions within the JEXL context:
```

docs/usermanual.md
```diff
@@ -843,6 +842,30 @@ the example case the command will be:
...
+
+-   `id`: device ID
+-   `type`: NGSI entity type (principal)
+-   `service`: device service (`Fiware-Service`)
+-   `subservice`: device subservice (`Fiware-ServicePath`)
+-   `staticAttributes`: static attributes defined in the device or config group
+-   `attribute`: attribute name and value involved in the command.
```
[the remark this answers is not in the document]
AlvaroVega: Done in https://github.com/telefonicaid/iotagent-json/pull/870/commits/399bcee817b9ddff321eb79b941c76efacf3207c

docs/usermanual.md
```diff
@@ -843,6 +842,38 @@ the example case the command will be:
...
+
+```json
+{"id":"disp2","type":"thing","service":"smartcity","subservice":"/","entity_name":"thing:disp2","color":"blue","brand":"tef","reset":0}
+```
+Where *color* and *brand* are static attrs and *reset* is the command name and *0* is the value for the command.
+
+This way an example for use jexl expressions for **endpoint** could be:
+```json
+{
+    "endpoint": "'https://myendpoint' + '/' + service + '/' + subservice + '/' + id + '/' + type"
```
pelayogb, line 866:
```suggestion
    "endpoint": "'https://myendpoint' + '/' + service + subservice + '/' + id + '/' + type"
```

docs/usermanual.md
```diff
@@ -843,6 +842,38 @@ the example case the command will be:
...
+-   `subservice`: device subservice (`Fiware-ServicePath`)
+-   `staticAttributes`: static attributes defined in the device or config group (flatted into context object).
+-   `attribute`: attribute name and value involved in the command.
+
+An example of JEXL context could be like:
+
+```json
+{"id":"disp2","type":"thing","service":"smartcity","subservice":"/","entity_name":"thing:disp2","color":"blue","brand":"tef","reset":0}
+```
+Where *color* and *brand* are static attrs and *reset* is the command name and *0* is the value for the command.
```
fgalan, line 861: I understand it is not a way of distinguish static attributes (color, brand) from the command involved in the command (reset). Is my understanding correct?

docs/usermanual.md
```diff
@@ -843,6 +842,38 @@ the example case the command will be:
...
+
+This way an example for use jexl expressions for **endpoint** could be:
+```json
+{
+    "endpoint": "'https://myendpoint' + '/' + service + subservice + '/' + id + '/' + type"
```
fgalan, line 866: I understand this is preferable over the previous alternative (`service + '/' + subservice `) given that subservice has the `/` char included. Is my understanding correct?

docs/usermanual.md
```diff
@@ -843,6 +842,38 @@ the example case the command will be:
...
+The agent provides additional information, hereafter referred as **JEXL context**, in order to be used to evaluate the
+expression of the **command** as well as the **device endpoint** when apply. In all cases the following data is available to all expressions within the JEXL context:
+
+-   `id`: device ID
+-   `type`: NGSI entity type (principal)
+-   `entity_name`: Entity Name
+-   `service`: device service (`Fiware-Service`)
+-   `subservice`: device subservice (`Fiware-ServicePath`)
+-   `staticAttributes`: static attributes defined in the device or config group (flatted into context object).
+-   `attribute`: attribute name and value involved in the command.
```
fgalan, line 854:
```suggestion
-  Static attributes defined in the device or config group (flatted into context object), by name.
-  The attribute involved in the command, by name.
```

I think is better not to use the `code style` in this case, because it is used by the others (`id`, etc.) and the reader may think they are literal references.

docs/usermanual.md
fgalan, line 861: Maybe a note on that can be added:

```suggestion
Where *color* and *brand* are static attrs and *reset* is the command name and *0* is the value for the command. Note that from a jexl context point of view, there is no distinction between static attributes and the attribute for the command (it is supposed that the writer of jexl expression knows the name of command and the names of the static attributes).
```

docs/usermanual.md
```diff
@@ -843,6 +842,38 @@ the example case the command will be:
...
+
+An example of JEXL context could be like:
+
+```json
+{"id":"disp2","type":"thing","service":"smartcity","subservice":"/","entity_name":"thing:disp2","color":"blue","brand":"tef","reset":0}
+```
+Where *color* and *brand* are static attrs and *reset* is the command name and *0* is the value for the command.
+
+This way an example for use jexl expressions for **endpoint** could be:
```
fgalan, line 863:
```suggestion
This way an example for use jexl expressions for **endpoint** could be (note that the `subservice` element includes the `/` as first character):
```

fgalan:
The PR looks ok, but something in the GitAction checks is failing...

fgalan (approved):
LGTM

